PR #2, opened 2021-10-13
PYI-307: Create skeleton app

kerrr:
A skeleton of the Orchestrator stub app. Displays a single page with GOV.UK branding.


di-ipv-orchestrator-stub/gradlew
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
merlinc, line 1: These Gradle scripts (`gradlew` & `gradle.bat`) look like boilerplate.

Are they up to date - and is there an alternate way of using them, apart from copying and pasting the repo?

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE html>
```
merlinc, line 1: If we're rendering templates, how much does this orchestrator need to be a Java application?

di-ipv-orchestrator-stub/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM gradle:7.2.0-jdk17@sha256:0619ec0c28cdd0f5f7ba389e7fea845b400cf282dd28b46b4d67f60a009732fa AS build
+COPY --chown=gradle:gradle . /home/gradle/src
+WORKDIR /home/gradle/src
+RUN gradle build --no-daemon
+
+FROM openjdk:17-jdk-slim@sha256:919c92bf200dc4faf73dcd858ac7a495edfafe7552e3a7ad6266c386a34bad60
+
+ENV PORT 8081
+RUN addgroup --system --gid 1001 appgroup && adduser --system --uid 1001 appuser --gid 1001
```
merlinc, line 9: Is `appuser` and `appgroup` our standard for permissions inside a container?

di-ipv-orchestrator-stub/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN mkdir /app
+RUN apt-get update && apt-get install -y curl
```
merlinc, line 11: I'm not sure about where I'd prefer this line to be. Normally it would be earlier on, as other commands are likely to invalidate this cache layer. However the previous commands are fairly static.

Any thoughts either way?

di-ipv-orchestrator-stub/README.MD
```diff
@@ -0,0 +1,30 @@
+# Digital Identity IPV Orchestrator Stub
+`di-ipv-orchestrator-stub`
+
+An Orchestrator stub using Spark Java
+
+## About
+
+THe starting point for manual testing and demonstrating IPV journeys.
```
merlinc:
```suggestion
The starting point for manual testing and demonstrating IPV journeys.
```

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/Orchestrator.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.di.ipv.stub.orc;
```
merlinc, line 1: This doesn't have the same path as the di-ipv-alpha-stubs` - what were the thoughts behind repackaging?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/Orchestrator.java
```diff
@@ -0,0 +1,22 @@
...
+
+import spark.Spark;
```
merlinc, line 3: The version of this in `di-ipv-alpha-stubs` is using static imports and they're not being used here - is there a specific reason for this difference?

di-ipv-orchestrator-stub/src/main/resources/public/gds/VERSION.txt
```diff
@@ -0,0 +1 @@
+3.11.0
```
merlinc, line 1: Is there a better way of using the GDS toolkit in a Java app, apart from copying-and-pasting a bunch of code?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/Orchestrator.java
```diff
@@ -0,0 +1,22 @@
...
+        Spark.staticFileLocation("/public");
+        Spark.port(Integer.parseInt(OrchestratorConfig.PORT));
+
+        initRoutes();
+    }
+
+    public void initRoutes(){
+        Spark.get("/", HomeHandler.serveHomePage);
+
+        Spark.internalServerError("<html><body><h1>Waaargh!!! Da Orc Boss sez we'ze got some gremlinz in da gearz.</h1></body></html>");
```
merlinc, line 19: I'm not usually a fan of "fun" messages like these (I've been on teams that have been burnt by them in the past) - but I do like the skull image you're using for a 500 message so 🤷

merlinc (commented):
Apologies for the rather scattered set of questions - I think I could have done with a pairing session / annotated PR to help with understanding the base decisions that help form this PR

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/Orchestrator.java
kerrr: I want the whole java codebase for IPV to exist in a well structured namespace, so that if it was all in one place there would be no clashes.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/Orchestrator.java
kerrr: This is better aligned with the GDS standards for Java. https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

di-ipv-orchestrator-stub/src/main/resources/public/gds/VERSION.txt
kerrr: Not that I'm aware of. I don't think it's a big issue for this stub.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/Orchestrator.java
kerrr: 😬  In general I'm not either. But maybe we can have a little fun in the Orc stub. Waaarrgghh!

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
kerrr: It doesn't really. The main advantage is that we can use Java client OAuth libs and that will probably be what the real Orchestrator will be using.

di-ipv-orchestrator-stub/gradlew
kerrr: This is the pattern that has been used for other Java apps in DI / Verify. It's a reasonable challenge though. We should establish if this is best practice and come back to this. For now I'm happy to follow convention.

di-ipv-orchestrator-stub/Dockerfile
merlinc: As discussed, these were just found in a previous commit

merlinc (approved):
As discussed offline, I would have preferred this to have been several distinct commit as that would have made it easier to review both now, and in the future.

That being said, this is for a piece of code that will not be deployed to production, so those sort of high standards can be saved for a future PR in a different repo

PR #13, opened 2021-11-25
[PYI-358] (PYI-273) - Implement token endpoint

tal-nagra:
### What changed
A new /token endpoint has been added to the credential issuer stub so that access tokens can be issued.
Error messages and input validation has been refactored to make further use of the nimbus jar classes.

### Why did it change
In order to issue an access token that can be used to access a protected resource.

### Issue tracking
- [PYI-358](https://govukverify.atlassian.net/browse/PYI-358)


idrop (approved):
I've seen this running and LGTM

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/AuthCodeService.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AuthCodeService {
+    private final Map<String, String> authCodes;
+
+    public AuthCodeService() {
+        this.authCodes = new ConcurrentHashMap<>();
+    }
+
+    public void persist(AuthorizationCode authCode, String resourcePayload) {
```
kerrr, line 15: I think this would be better if the authCode was a sting here, making the `put` and `get` symmetrical.

kerrr: But not a blocker.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/TokenService.java
```diff
@@ -0,0 +1,33 @@
...
+public class TokenService {
+
+    private static final long DEFAULT_ACCESS_TOKEN_TTL_IN_SECS = 3600;
+    private final Map<String, String> accessTokens;
+
+    public TokenService() {
+        this.accessTokens = new ConcurrentHashMap<>();
+    }
+
+    public void persist(AccessToken accessToken, String resourcePayload) {
```
kerrr, line 18: Same as the AuthCode equivalent. I think this would be better if the accessToken was a sting here, making the `put` and `get` symmetrical.

Not a blocker. If it causes any issues we can change it.

kerrr (approved):
🚀  Onwards!

PR #17, opened 2021-12-02
PYI-300/ get gradle build to work on M1

SunilDGDS:
### What changed

The 2 docker files have been updated to use gradle 7.3 instead of 7.2  that was stopping the build esp on Apple M1

- [PYI-300](https://govukverify.atlassian.net/browse/PYI-300)

### DIFF
--- a/di-ipv-orchestrator-stub/Dockerfile
+++ b/di-ipv-orchestrator-stub/Dockerfile
@@ -1,4 +1,4 @@
-FROM gradle:7.2.0-jdk17@sha256:0619ec0c28cdd0f5f7ba389e7fea845b400cf282dd28b46b4d67f60a009732fa AS build
+FROM gradle:7.3.0-jdk17@sha256:1f65f4991aa496cf835cbce10290bdc71c1cb526b4df807d97c4f51b863f2399 AS build
 COPY --chown=gradle:gradle . /home/gradle/src
 WORKDIR /home/gradle/src
 RUN gradle build --no-daemon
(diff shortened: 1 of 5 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-bin.zip
```
jamietanna, line 3: If we're targeting 7.3, we should be doing this:

```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
```

di-ipv-orchestrator-stub/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM gradle:7.2.0-jdk17@sha256:0619ec0c28cdd0f5f7ba389e7fea845b400cf282dd28b46b4d67f60a009732fa AS build
+FROM gradle:7.3.0-jdk17@sha256:1f65f4991aa496cf835cbce10290bdc71c1cb526b4df807d97c4f51b863f2399 AS build
```
jamietanna, line 1: An alternative way of doing this would be to just use a JDK image, and then use `./gradlew`, which means we don't need to tie our Gradle version to the version in the Wrapper - thoughts?

kerrr:
We should have a discussion about builds across the whole of stubs. The repo was set up to just be a bucket for completely separate projects that existed in their own directories. That might be a terrible way of doing it, but we should discuss changing that model if that's what we wan to do.

@SunilDGDS For now can you modify this PR to 
 * not add the 3 new files
 * update the gradle wrapper in `di-ipv-orchestrator-stub`
 * update the gradle wrapper in `di-ipv-credential-issuer-stub`

di-ipv-orchestrator-stub/Dockerfile
kerrr: Yes, that might be a good way of doing it.

PR #20, opened 2022-01-06
BAU: Bump gradle image version to 7.3.3 and remove SHAs

Wynndow:
### Why did it change
The versions of the gradle and openjdk images being referred to with
their SHAs were for amd64 architecture. These images didn't work on an
M1 which is arm64.

As we now have a mix of architectures being used for development we
should remove these SHAs and let Docker decided which version of the
image to pull.

This also stops using a gradle image and instead uses the openjdk image
with gradle wrapper. This should be easier to maintain going forwards.


danworth (approved):
Looks good 👍  I can build and run each of the updated images (I haven't switched to M1 mac yet). I guess removing the SHAs removes some element of protection against supply chain attacks/ things changing without us noticing since tags are mutable but these are stubs for local testing and they're well known and supported images, plus I don't know of an alternative way of dealing with different architectures in the interim.

PR #32, opened 2022-01-31
PYI-618: Load client signing certificate based off provided client id

dannyhayes-gds:
### What changed
Load the client config env variable and parse into JSON.

### Why did it change
This config variable will be used to determine which signing cert is required per client.

### Issue tracking

- [PYI-618](https://govukverify.atlassian.net/browse/PYI-618)


SunilDGDS (changes requested):
This is only a general comment .. I think it's best to use nimbus StringUtils.IsBlank when checking for null values ; this utility checks for empty & null

SunilDGDS (approved):

PR #33, opened 2022-02-01
PYI-575: Validate jwt signature

dannyhayes-gds:
### What changed

- Get the public key of the client cert
- Use key to verify the signature of the JWT
- Update validation and error handling of missing or invalid JWT/client config

### Why did it change
To allow the cri stub to accomodate multiple clients and be able to verify the signature of any JWT's by using its singing cert provided in the cri configuration.

### Issue tracking

- [PYI-575](https://govukverify.atlassian.net/browse/PYI-575)

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
+++ b/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
@@ -4,6 +4,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.crypto.RSASSAVerifier;
 import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.AuthorizationCode;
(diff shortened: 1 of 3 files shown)


galund:
I'm a bit confused here by the term "client". In this transaction, IPV is the client, and a credential issuer is the "resource server" (to use the OAuth terminology.

Caveat: it could be I'm wrong about what you're up to, but if so I will learn something :-)

dannyhayes-gds:
You are correct, in reality the only client we will most likely have (at least in the short term) is IPV core. But we wanted to implement this in a way that the cri could handle multiple clients if ever required, this includes if we do anything with rolling out new IPV core certificates and want to use them under a new ipv core client ID etc.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
```diff
@@ -237,22 +246,46 @@ private String getSharedAttributes(QueryParamsMap queryParamsMap) {
...
     }
+
+    private boolean isInvalidSignature(SignedJWT signedJWT, Map<String, String> client)
+            throws CertificateException, JOSEException {
+        byte[] binaryCertificate =
+                Base64.getDecoder().decode(client.get(SIGNING_CERT_PARAM));
+        CertificateFactory factory = CertificateFactory.getInstance("X.509");
+        Certificate certificate = factory.generateCertificate(new ByteArrayInputStream(binaryCertificate));
+
+        PublicKey publicKey = certificate.getPublicKey();
```
idrop, line 285: Just an observation: if we're just using the x509 cert as a way of storing the public key, can we just have the public key in config instead? If there's intent to validate the expiry and chain of the cert later on, then I guess this makes sense.

Wynndow: We had a bit of a chat about this and decided on going down the X509 route. Mostly because it'll be consistent with what we're doing elsewhere, so in a few months we don't go "oh, why did we do this differently?".

galund: There's probably a discussion to have at some point about whether a "current set" of public keys could be loaded into configuration somewhere, so that we don't have to be parsing certificates all over the place on every request, but I'm all for not prematurely optimising.

eg we might well want a check for revocation when we check the cert/chain, but not necessarily on every single request.

(And cc @philandstuff as it's more his wheelhouse than mine!)

kerrr (approved):
LGTM

PR #36, opened 2022-02-15
PYI-695: Validate redirect urls

Wynndow:
#### Add redirect URL validator function

Add a function to the Validator to compare the redirect URL received in
the query parameters to those configured in the client configuration.

Refactors the CredentialIssuersConfig class to parse the `CLIENT_CONFIG`
env variable and deserialize individual configs into objects. This was
necessary as we now have tested config and it was getting horrible messy
trying to splunk into it.

#### Validate redirect URI in token endpoint

Validate that the redirect URI included in the grant on the token
endpoint matches that received in the auth endpoint.

### Issue tracking

- [PYI-695](https://govukverify.atlassian.net/browse/PYI-695)


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/config/ClientConfig.java
```diff
@@ -0,0 +1,16 @@
...
+public class ClientConfig {
+    public String getSigningCert() {
+        return signingCert;
+    }
+
+    public Map<String, String> getJwtAuthentication() {
+        return jwtAuthentication;
+    }
+
+    private String signingCert;
```
dannyhayes-gds: Could we move these vars to the top of the file, think we've always got the getter's/setter methods below the variable elsewhere

SunilDGDS (dismissed):
Looks good

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
```diff
@@ -209,43 +212,37 @@ private void persistData(QueryParamsMap queryParamsMap, AuthorizationCode author
...
             return new ValidationResult(false, OAuth2Error.UNSUPPORTED_RESPONSE_TYPE);
         }
 
         String clientIdValue = queryParams.value(RequestParamConstants.CLIENT_ID);
-        if (Validator.isNullBlankOrEmpty(clientIdValue)) {
+        if (Validator.isNullBlankOrEmpty(clientIdValue) || CredentialIssuerConfig.getClientConfig(clientIdValue) == null) {
             return new ValidationResult(false, OAuth2Error.INVALID_CLIENT);
         }
 
+        if (Validator.redirectUrlIsInvalid(queryParams)) {
```
dannyhayes-gds: It could be worth adding an additional check above this to make sure the redirect_uri has been provided. And that can return its own error of something along the lines of a "Missing redirect uri param" error.

And then this check can specifically be just checking that the provided redirect_uri is valid or not according to our config

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
```diff
@@ -209,43 +212,37 @@ private void persistData(QueryParamsMap queryParamsMap, AuthorizationCode author
...
+            return new ValidationResult(false, new ErrorObject(
+                    ERROR_CODE_INVALID_REDIRECT_URI,
+                    "redirect_uri param must be provided",
```
dannyhayes-gds, line 230: If you agree with the above then this error description should maybe be changed slightly to be specifcally something like 

```suggestion
                    "redirect_uri param provided does not match any of the allowed redirect_uri values configured",
```

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
```diff
@@ -209,43 +212,37 @@ private void persistData(QueryParamsMap queryParamsMap, AuthorizationCode author
...
+
         return ValidationResult.createValidResult();
     }
 
     private String getSharedAttributes(QueryParamsMap queryParamsMap) {
         String requestParam = queryParamsMap.value(RequestParamConstants.REQUEST);
         String clientIdParam = queryParamsMap.value(RequestParamConstants.CLIENT_ID);
 
-        if (StringUtils.isBlank(CredentialIssuerConfig.CLIENT_CONFIG)) {
+        if (CredentialIssuerConfig.CLIENT_CONFIGS.isEmpty()) {
```
dannyhayes-gds: Slightly concerned about this bit and specifically when the CLIENT_CONFIGS var gets set with a value. As it starts out as null and only gets a value after the 1st call to CredentialIssuerConfig.getClientConfig() I think? Not sure if there is a danger of a null pointer here?

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/AuthCodeService.java
```diff
@@ -1,23 +1,38 @@
...
-        this.authCodes.put(authCode.getValue(), resourcePayload);
+    public void persist(AuthorizationCode authCode, String resourcePayload, String redirectUrl) {
+        this.authCodes.put(authCode.getValue(), Map.of(RESOURCE_PAYLOAD, resourcePayload, REDIRECT_URL, redirectUrl));
     }
 
     public String getPayload(String authCode) {
-        return this.authCodes.get(authCode);
+        try {
+            return this.authCodes.get(authCode).get(RESOURCE_PAYLOAD);
+        } catch (NullPointerException e) {
```
dannyhayes-gds: Generally I'd avoid trying to catch NullPointerException instead of testing against null values before using it. But I guess we don't need to be as strict with this being a Stub so fine to leave as is but might be a better alternative to test this.authCodes.get(authCode) is null or not first before trying to get the resource payload?

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
Wynndow: I think technically not providing a redirect_uri is valid, so I didn't want to enforce it here. I've got a ticket to look at if we can support it. If a client only has one redirect_url registered, they should be able to skip providing one as it's redundant.
Having said that, this validation doesn't actually handle that situation.
I'll add the check now, and then address that situation if we get to that ticket.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
Wynndow: yeah, good shout. I'll update it to something null safe.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/AuthCodeService.java
Wynndow: Yep, fair. I'll test for null then return.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
Wynndow: Having looked at the error objects provided by nimbus, I think I'm going to change these to use a standard one. Maybe add some logging to make it easier to work out what went wrong.

Wynndow: Actually, perhaps not...

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/config/CredentialIssuerConfig.java
```diff
@@ -1,20 +1,43 @@
...
     public static final String NAME = getConfigValue("CREDENTIAL_ISSUER_NAME","Credential Issuer Stub");
-    public static final String CLIENT_CONFIG = getConfigValue("CLIENT_CONFIG", null);
 
     public static final String EVIDENCE_STRENGTH_PARAM = "strength";
     public static final String EVIDENCE_VALIDITY_PARAM = "validity";
     public static final String ACTIVITY_PARAM = "activity";
     public static final String FRAUD_PARAM = "fraud";
     public static final String VERIFICATION_PARAM = "verification";
 
+    public static Map<String, ClientConfig> CLIENT_CONFIGS;
```
dannyhayes-gds, line 22: Is there a reason we can't set a value this straight away? As at the moment it will work fine, but when we are reading this value we are dependant that getClientConfig(clientId) has been called for it to have a value, so if we ever do any refactors etc we could accidentally end up trying to access this var before that method call and get an incorrect scenario where it looks like the client configs aren't setup.

Could this be moved below line 25 to be something like 
```suggestion
    public static Map<String, ClientConfig> CLIENT_CONFIGS = parseClientConfigs();
```

Wynndow: I did this because it makes the tests work! There might be a better way to do it, but I'm tempted to leave it for now.

dannyhayes-gds (approved):

PR #37, opened 2022-02-16
BAU: Update env var name to match other naming patters for ttl vars

dannyhayes-gds:
### What changed
Update the env var name for the orc stub client jwt ttl value and swap to seconds instead of minutes for the value.

### Why did it change
Match the same naming pattern as ipv-core and the cri's.


Wynndow (dismissed):

Danomanic (dismissed):

Danomanic (dismissed):
Looks good - do you want to squash the commits before merge?

dannyhayes-gds:
Ah yeah good shout, I'll squash them

Danomanic (approved):

PR #38, opened 2022-02-16
PYI-698: Accept JWT client authentication

Wynndow:
**Config PR to support these changes: https://github.com/alphagov/di-ipv-config/pull/163**

### What changed

Adds a new service to validate client authentication JWTs in the token
endpoint.

Also refactors some services and injects them to make things easier to
test.

### Issue tracking

- [PYI-698](https://govukverify.atlassian.net/browse/PYI-698)


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/auth/ClientJwtVerifier.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        try {
+            verifier.verify(authenticationJwt, null, null);
+        } catch (InvalidClientException | JOSEException e) {
+            throw new ClientAuthenticationException(e);
+        }
+    }
+
+    private Map<String, List<String>> listifyParamValues(QueryParamsMap requestParams) {
```
dannyhayes-gds: We might not need this method. The PrivateKeyJWT.parse(queryParams) has a few different overloaded methods. One of them accepts just a single string value which are the query params i.e "var1=hello&var2=bye&var=hello" etc...

So maybe instead of passing in the QueryMap object into the authenticateClient method, we can instead pass in the original query params as a string, which I think we can get from something like
request.queryString() 

Just a suggestion though, not too fussed if we want to keep it as is and build the map into the required Map<String, List<String>> format

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/auth/ClientJwtVerifier.java
```diff
@@ -0,0 +1,71 @@
...
+        if (clientConfig == null) {
+            throw new ClientAuthenticationException(String.format("Config for client ID '%s' not found", authenticationJwt.getClientID().getValue()));
+        }
+
+        if (clientConfig.getJwtAuthentication().get(AUTHENTICATION_METHOD).equals(NONE)) {
+            return;
+        }
+
+        try {
+            verifier.verify(authenticationJwt, null, null);
```
dannyhayes-gds, line 51: Do you know if this will test against the expiry ttl time being too far in the future? Or do we need to do that separately?

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/auth/ClientJwtVerifier.java
Wynndow: Nice. Have pushed a commit to use the query string version instead.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/auth/ClientJwtVerifier.java
Wynndow: I don't think it does. I think we agreed that for the stub we're not too concerned about this. I've spoken to @SunilDGDS about having to do it as an extra check in his core-back verifying PR though.

dannyhayes-gds: Ah okay thats sounds good 👍

dannyhayes-gds (approved):
Looks good 👍

PR #39, opened 2022-02-17
BAU: Add GitHub checks

Wynndow:
### What changed

* Create a root gradle project to run builds and checks from.
* Add a GitHub workflow to run the checks
* Add spotless to the root project
* Run spotlessApply

### Why did it change

We use spotless in our other Java projects to keep code consistent. We
should do this too.

The exports in the gradle.properties files are necessary as we use Java
17 in the stubs. See this GitHub issue more the lo-down:
https://github.com/diffplug/spotless/issues/834#issuecomment-819118761

The first commit are the changes to add the GitHub checks and spotless.
The second commit is just running spotless against the project.


danworth (approved):
Looks good, thanks for adding this in 🏅

PR #42, opened 2022-02-21
BAU: Commit gradle-wrapper.jar

Wynndow:
### Why did it change

Without committing the jar the GitHub checks can't run.


danworth (approved):
Action now runs ok 👍

PR #43, opened 2022-02-21
PYI-677 Relax Redirect Url Validation on Stubs

danworth:
To enable the simpler reusing of stubs across Developer environments
accept redirect urls that include the London PaaS domain. The validation
could be tighter since this will accept a URI where the PaaS domain
simply appears in the path but these are stubs and including the extra
logic to guard against this seems unnecessary at present.

## Proposed changes
Validate that the redirect uri includes the London PaaS domain rather than a domain specific to a particular core front app. This allows them to be reused without needing to configure the stubs to know about every developer environment.


Danomanic (dismissed):
Looks good!

Danomanic (dismissed):

Danomanic (approved):

PR #45, opened 2022-02-21
Fix query params being passed into client authentication method

dannyhayes-gds:
### What changed
Refactored how the form-url encoded body is passed into the client authenticate method.

- Use the request.queryMap() method
- Convert QueryParamsMap into required Map<String, List<String>> structure 

### Why did it change
request.queryString() doesn't return form-url-encoded body values so the request.queryMap() needs to be used instead.


danworth (approved):
Sounds sensible hopefully solves the issue we were seeing regarding the `Missing client_assertion_type parameter` 👍

PR #51, opened 2022-03-03
PYI-731: Return Verified Credential from CRI stub

Wynndow:
### What changed

Returns a Verified Credential from CRI stub in JWT format. See here for
details: https://www.w3.org/TR/vc-data-model/#json-web-token

As this is this stub we're just returning all of the shared attributes
received. These should be in the format defined in our schema. It can be
found here: https://vocab.london.cloudapps.digital/

We're also including any data sent in via the form on the stubs
frontend.

The JWK is signed using an EC key.

This also DRYs up some of the tests.

### Issue tracking

- [PYI-731](https://govukverify.atlassian.net/browse/PYI-731)


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/CredentialIssuer.java
```diff
@@ -45,7 +44,7 @@ private void initRoutes() {
         Spark.get("/authorize", authorizeHandler.doAuthorize);
         Spark.post("/authorize", authorizeHandler.generateResponse);
         Spark.post("/token", tokenHandler.issueAccessToken);
-        Spark.get("/credential", credentialHandler.getResource);
+        Spark.post("/credentials/issue", credentialHandler.getResource);
```
dannyhayes-gds, line 47: You will need to update core-back parameter store values to use this new endpoint I think. For the new developer environments I think that will be here: https://github.com/alphagov/di-ipv-config/blob/main/core/parameters/dev.parameters.json#L31

Wynndow: Yep, there’s another story to make the changes to core back. I’d hold off merging this till that’s ready to save breaking things.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
```diff
@@ -151,6 +152,11 @@ public AuthorizeHandler(
                     Map<String, Object> attributesMap =
                             generateJsonPayload(queryParamsMap.value(JSON_PAYLOAD_PARAM));
 
+                    Map<String, Object> combinedAttributeJson =
+                            generateJsonPayload(getSharedAttributes(queryParamsMap));
```
dannyhayes-gds, line 156: Why do we need to return the shared attributes in the VC? I always assumed they would just be a separate thing since we have a lambda that goes and generates them upon each request to a CRI?

Wynndow: I did it to demonstrate that a cri can use the shared attributes when generating a vc and include them. We could return anything from the stub though really.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/vc/VerifiableCredentialGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private SignedJWT generateAndSignVerifiableCredentialJwt(String subject, Map<String, Object> vc)
+            throws NoSuchAlgorithmException, InvalidKeySpecException, JOSEException {
+        Instant now = Instant.now();
+        JWTClaimsSet claimsSet =
+                new JWTClaimsSet.Builder()
+                        .claim(SUBJECT, subject)
+                        .claim(ISSUER, System.getenv(VC_ISSUER_ENV_VAR))
+                        .claim(NOT_BEFORE, now.getEpochSecond())
+                        .claim(EXPIRATION_TIME, now.plusSeconds(60 * 5).getEpochSecond())
```
dannyhayes-gds: Is this going to be enough time for the VC? Perhaps it could be an env var so that we can control this?

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
dannyhayes-gds: Ah okay fair enough, so its not something we would be doing for real CRI's? If so then that sounds all good 👍

Wynndow: Yeah, I think it’s dependent on the real CRIs needs what they do with them.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/vc/VerifiableCredentialGenerator.java
Wynndow: Ah yeah, that’s a good shout. George thought 5 mins would be ok for the real thing, but it would be good for it to be configurable.

Wynndow: @dannyhayes-gds Have pushed a commit to grab a value from the env (as well as moving other config stuff to the existing config class). There's also a PR here to set a value in the env from the manifest: https://github.com/alphagov/di-ipv-config/pull/207

dannyhayes-gds (approved):
Looks good 👍

PR #53, opened 2022-03-08
KBV-87 pick up core-stub CRI config from a file deployed with the app

idrop:
## Proposed changes

Set CRI config from a file in the app rather than an env var.

This means we don't have to fiddle with base64 encoding yaml files to env vars.

The path to the cri-dev.yaml file for the PaaS deploy has already been set with:
````
cf set-env di-ipv-core-stub CORE_STUB_CONFIG_FILE config/cris-dev.yaml
````
I've tested this on Docker with `./startup.sh`.

### DIFF
--- a/di-ipv-core-stub/README.MD
+++ b/di-ipv-core-stub/README.MD
@@ -14,6 +14,12 @@ This stub allows us to:
 * send these claims to a credential issuer
 * view the credentials returned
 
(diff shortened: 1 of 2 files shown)


MattBidewell (approved):
Code looks good and using a config file should make it easier with testing! 😁

PR #60, opened 2022-03-28
Add CORE_STUB_KEYSTORE_FILE property

merlinc:
### What changed

- Add CORE_STUB_KEYSTORE_FILE property
- Add conditional logic to handle missing both of keystore_file and keystore_base64 properties

### Why did it change

This Docker image needs to be built without config in it to be a publicly available image

### Issue tracking

- [KBV-254](https://govukverify.atlassian.net/browse/KBV-254)

### Other considerations


di-ipv-core-stub/Dockerfile
```diff
@@ -21,7 +21,6 @@ COPY --from=build \
 	/app/
 
 WORKDIR /app
-ADD config config
```
MattBidewell, line 24: ~~Why are you removing the adding of the config directory?~~

Edit: I reread the PR. Its so it can be public facing?

MattBidewell (dismissed):

di-ipv-core-stub/Dockerfile
merlinc: I'm not that fussed about whether the image is public or private - I just need it somewhere that it can be easily accessed. The config on the other hand is definitely private, and I don't want an entire config repo exposed for a few settings.

So the approach is build the image so it works without config. Locally it can be run with a sibling path to the config. In a PR we get the necessary config from a S3 bucket, which reduces the blast radius if something went wrong.

di-ipv-core-stub/compose.yaml
```diff
@@ -0,0 +1,17 @@
+version: "3.9"
```
merlinc, line 1: The script & .env file have been swapped for a docker-compose file with environment variables and volumes. The secret stuff is still elsewhere

di-ipv-core-stub/README.MD
```diff
@@ -17,29 +17,30 @@ This stub allows us to:
 ## Config
 
 Core Stub config from the `di-ipv-config` repository directory `/stubs/di-ipv-core-stub` will be available at:
-* Docker image: `/app/config/`
+* Docker image: `/app/`
```
merlinc, line 20: Rather than create an empty config directory, it seemed easier to mount the required files into the existing app directory.

MattBidewell (dismissed):
Looks good to me

My understanding 
- We're providing the config variables as a file now, file loc to be set in the .env. 

My only suggestion would be provide an example env file, if needed.

merlinc:
> Looks good to me
> 
> My understanding
> 
> * We're providing the config variables as a file now, file loc to be set in the .env.

Yeah, there's an additional property now for the "keystore as a keystore file" to match the "keystore as a base64 string".

> My only suggestion would be provide an example env file, if needed.

Local usage of the stub doesn't change. The config directory is referenced by the Dockerfile, and that's mentioned in the `.env.sample` file as well.

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/CoreStub.java
```diff
@@ -58,11 +59,25 @@ private ExceptionHandler exceptionHandler() {
...
-                        Base64.getDecoder().decode(CoreStubConfig.CORE_STUB_KEYSTORE_BASE64))) {
-            keystore.load(inputStream, CoreStubConfig.CORE_STUB_KEYSTORE_PASSWORD.toCharArray());
+
+        if (CoreStubConfig.CORE_STUB_KEYSTORE_FILE != null) {
+            try (FileInputStream inputStream =
+                    new FileInputStream(CoreStubConfig.CORE_STUB_KEYSTORE_FILE)) {
+                keystore.load(
+                        inputStream, CoreStubConfig.CORE_STUB_KEYSTORE_PASSWORD.toCharArray());
+            }
+        } else if (CoreStubConfig.CORE_STUB_KEYSTORE_BASE64 != null) {
```
MattBidewell, line 65: Heads up but looking at the current deployment in PaaS, this has already been set. So you might need to update it

MattBidewell (dismissed):
lgtm - just to note a the base64 keystore env has been added in paas already, you might need to update it or something

PR #62, opened 2022-03-30
KBV-231: Changes for the custom claims structure

techaman-m:
A specification was required for how data is passed in from IPV Core to credential issuers so that both IPV Core and the credential issuer team(s) can track their compliance against more easily.

### What changed

Refactored the classes and objects to create a custom claim object called `shared_claims` as discussed in meetings with @galund and tech leads.

### Current Format:

```{
  "sub": "urn:uuid:2328d856-524f-431e-a7e4-acf0168c9f7d",
  "aud": "https://experian.cri.account.gov.uk",
  "nbf": 1648717302,
  "shared_claims": {
    "name": [
      {
        "nameParts": [
          {
            "type": "GivenName",
            "value": "xxxxxx"
          },
          {
            "type": "FamilyName",
            "value": "xxxxxx"
          }
        ]
      }
    ],
    "@context": [
      "https://www.w3.org/2018/credentials/v1",
      "https://vocab.london.cloudapps.digital/contexts/identity-v1.jsonld"
    ],
    "birthDate": [
      {
        "value": "1970-01-01"
      }
    ]
  },
  "iss": "https://dev.core.ipv.account.gov.uk",
  "exp": 1648720902,
  "iat": 1648717302
}
```

### Issue tracking
[Related RFC](https://github.com/alphagov/digital-identity-architecture/pull/109)

- [KBV-231](https://govukverify.atlassian.net/browse/KBV-231)

We can consider adding these as environment variables on PaaS:

CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI=https://experian.cri.account.gov.uk
CORE_STUB_JWT_ISS_CRI_URI=https://dev.core.ipv.account.gov.uk

### Other considerations

- The structure can change based on further custom claims which may be introduced at a later stage for other credential issuers such as address, passport etc. Currently it has been formulated to work with the KBV CRI
- The structure can also change to contain information related to documents that can be used for DVLA and Passport flows and this is all for future consideration

### DIFF
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
@@ -48,6 +48,11 @@ public class CoreStubConfig {
     public static final String CORE_STUB_KEYSTORE_ALIAS =
             getConfigValue("CORE_STUB_KEYSTORE_ALIAS", null);
 
+    public static final String CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI = getConfigValue("CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI", "https://experian.cri.account.gov.uk");
+    public static final String CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI = getConfigValue("CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI", "https://ipv.core.account.gov.uk");
+    public static final String CORE_STUB_CONTEXT = getConfigValue("CORE_STUB_CONTEXT", "https://www.w3.org/2018/credentials/v1");
+    public static final String CORE_STUB_JSON_SCHEMA = getConfigValue("CORE_STUB_JSON_SCHEMA", "https://vocab.london.cloudapps.digital/contexts/identity-v1.jsonld");
+
     public static final List<Identity> identities = new ArrayList<>();
     public static final List<CredentialIssuer> credentialIssuers = new ArrayList<>();
 
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
@@ -51,14 +51,8 @@ public class CoreStubConfig {
     public static final String CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI =
             getConfigValue(
                     "CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI", "https://experian.cri.account.gov.uk");
-    public static final String CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI =
-            getConfigValue("CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI", "https://ipv.core.account.gov.uk");
-    public static final String CORE_STUB_CONTEXT =
-            getConfigValue("CORE_STUB_CONTEXT", "https://www.w3.org/2018/credentials/v1");
-    public static final String CORE_STUB_JSON_SCHEMA =
-            getConfigValue(
-                    "CORE_STUB_JSON_SCHEMA",
-                    "https://vocab.london.cloudapps.digital/contexts/identity-v1.jsonld");
+    public static final String CORE_STUB_JWT_ISS_CRI_URI =
+            getConfigValue("CORE_STUB_JWT_ISS_CRI_URI", "https://dev.core.ipv.account.gov.uk");
 
     public static final List<Identity> identities = new ArrayList<>();
     public static final List<CredentialIssuer> credentialIssuers = new ArrayList<>();
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
@@ -48,10 +48,17 @@ public class CoreStubConfig {
     public static final String CORE_STUB_KEYSTORE_ALIAS =
             getConfigValue("CORE_STUB_KEYSTORE_ALIAS", null);
 
-    public static final String CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI = getConfigValue("CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI", "https://experian.cri.account.gov.uk");
-    public static final String CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI = getConfigValue("CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI", "https://ipv.core.account.gov.uk");
-    public static final String CORE_STUB_CONTEXT = getConfigValue("CORE_STUB_CONTEXT", "https://www.w3.org/2018/credentials/v1");
-    public static final String CORE_STUB_JSON_SCHEMA = getConfigValue("CORE_STUB_JSON_SCHEMA", "https://vocab.london.cloudapps.digital/contexts/identity-v1.jsonld");
+    public static final String CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI =
+            getConfigValue(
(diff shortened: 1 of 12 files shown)


di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
```diff
@@ -48,6 +48,18 @@
     public static final String CORE_STUB_KEYSTORE_ALIAS =
             getConfigValue("CORE_STUB_KEYSTORE_ALIAS", null);
 
+    public static final String CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI =
+            getConfigValue(
+                    "CORE_STUB_JWT_AUD_EXPERIAN_CRI_URI", "https://experian.cri.account.gov.uk");
+    public static final String CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI =
+            getConfigValue("CORE_STUB_JWT_ISS_EXPERIAN_CRI_URI", "https://ipv.core.account.gov.uk");
```
galund: There's something funny about this config name - the `iss` should be the same regardless of what CRI we're calling

galund: Also it doesn't seem right to have a default that looks like the prod name, in the stub. have a chat with @idrop about this one maybe

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
```diff
@@ -48,6 +48,18 @@
...
+    public static final String CORE_STUB_CONTEXT =
+            getConfigValue("CORE_STUB_CONTEXT", "https://www.w3.org/2018/credentials/v1");
+    public static final String CORE_STUB_JSON_SCHEMA =
```
galund: This isn't a JSON Schema, it's another JSON-LD "@context" value - which is a list. I think potentially you just want a template JSON file here - neither of these context values needs to be coming out of config. It won't change per environment.

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/FindDateOfBirth.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.di.ipv.stub.core.config.uatuser;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Date;
+
+public record FindDateOfBirth(Instant dateOfBirth, Instant dateOfEntryOnCtdb) {
+
+    public Date getDOB() {
```
galund, line 9: This function seems really weird, is it just a way of converting from one date format to another? If so, be explicit about what you're doing rather than doing an implied conversion?

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
```diff
@@ -47,9 +52,9 @@ public Identity map(Map<String, String> map, int rowNumber) {
 
         Instant dob = Instant.parse(map.get("dob"));
         Instant dateOfEntryOnCtdb = Instant.parse(map.get("dateOfEntryOnCtdb"));
-        DateOfBirth dateOfBirth = new DateOfBirth(dob, dateOfEntryOnCtdb);
+        FindDateOfBirth dateOfBirth = new FindDateOfBirth(dob, dateOfEntryOnCtdb);
 
-        Name name = new Name(map.get("name"), map.get("surname"));
+        FullName name = new FullName(map.get("name"), map.get("surname"));
```
galund, line 55: I don't think you're using this - what's it for?

galund (commented):
Couple of suggestions here but nothing major

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/FindDateOfBirth.java
techaman-m: This function is specific to the way we calculate the DOB for the Experian KBV dataset. Each identity has 2 date values that allows us to calculate the actual date of birth for the identity. 

To calculate the DOB for an identity we need to apply the following logic : 
`todays date - date of entry + date of birth`

So technically this object is helping us calculate/find the actual DOB for the identity in the Experian Dataset. This is specific to our KBV stub and is like a helper Object to get the DOB. 

I have kept the `DateOfBirth` object which is part of the `shared_claims` separate so as to allow separation of concerns. We can always remove the `FindDateOfBirth` or change it based on use case in future but the actual date of birth will always be fed into the `DateOfBirth` object in the `shared_claims`.

This is the `shared_claims` object [here](https://github.com/alphagov/di-ipv-stubs/blob/73ce83aaf81686504df94da51d5d81cd30a2f52d/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/SharedClaims.java)

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
techaman-m: That is right, I am not using it in the `shared_claims` object

Earlier `Name` was being used here but its actual purpose was to get the stub data from Experian. This is again specific to our KBV dataset from which we are picking up the name of the identity. I felt that `Name` should be part of the `shared_claims`, hence shuffling it around to keep objects in `shared_claims` as readable as possible

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
techaman-m: Agreed, I have renamed that to `CORE_STUB_JWT_ISS_CRI_URI`

techaman-m: I have renamed it to `https://dev.core.ipv.account.gov.uk` and also asked @idrop to leave a comment

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
techaman-m: Agreed, I have put it back to strings as I dont think we have the JSON-LD yet

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/CoreStubConfig.java
idrop: 👍 makes sense to me @amanm-alphagov

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/FindDateOfBirth.java
idrop: @galund For whatever reason, the Experian test users keep their age e.g. perpetually 40 years old. This fn models that.

idrop:
I've just done a test on this branch and the JWT payload looks like:
````
{
  "sub": "urn:uuid:724afc7d-d8a2-45ba-888c-7e3b57db5877",
  "aud": "https://experian.cri.account.gov.uk",
  "nbf": 1648799845,
  "shared_claims": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1",
      "https://vocab.london.cloudapps.digital/contexts/identity-v1.jsonld"
    ],
    "name": [
      {
        "nameParts": [
          {
            "type": "GivenName",
            "value": "xxxxx"
          },
          {
            "type": "FamilyName",
            "value": "xxxxx"
          }
        ]
      }
    ],
    "birthDate": [
      {
        "value": "1978-09-08"
      }
    ]
  },
  "iss": "https://dev.core.ipv.account.gov.uk",
  "exp": 1648803445,
  "iat": 1648799845
}
````
This looks ok to me, so I'm approving

idrop (approved):

PR #75, opened 2022-04-24
KBV-300 set IPV Core Stub aud from config

idrop:
### What changed

Set the IPV Core Stub audience from ipv-config repo

### DIFF
--- a/di-ipv-core-stub/Dockerfile
+++ b/di-ipv-core-stub/Dockerfile
@@ -1,11 +1,11 @@
 FROM openjdk:17-jdk-slim AS build
 
 WORKDIR /home/gradle/src
-COPY src src
 COPY gradlew .
 COPY gradle gradle
 COPY build.gradle .
-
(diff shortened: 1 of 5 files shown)


di-ipv-core-stub/Dockerfile
```diff
@@ -1,11 +1,11 @@
...
 
 WORKDIR /home/gradle/src
-COPY src src
 COPY gradlew .
 COPY gradle gradle
 COPY build.gradle .
-
+RUN ./gradlew build --no-daemon
+COPY src src
 RUN ./gradlew build --no-daemon
```
idrop, line 9: run a separate gradle build on `src` only - avoids long gradle dep builds in the case where only src changes

MattBidewell (dismissed):
Makes sense to me, also makes sense why we where bashing our heads on Friday with the audience issue

KaraokeStu (approved):
LGTM

PR #89, opened 2022-05-10
PYIC-969 Update orc stub to sign and encrypt a new request jwt

blakeyp:
Co-authored by @AmritSidhu 

### What changed

Orchestrator to send signed+encrypted JAR request to IPV core

### Why did it change

More secure version of auth request and matches what real orchestrator will be sending

### Issue tracking

- [PYIC-969](https://govukverify.atlassian.net/browse/PYIC-969)

- [X] Added to deployment repository

### DIFF
--- a/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
+++ b/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
@@ -16,8 +16,12 @@ public class OrchestratorConfig {
             getConfigValue("ORCHESTRATOR_CLIENT_ID", "di-ipv-orchestrator-stub");
     public static final String ORCHESTRATOR_REDIRECT_URL =
             getConfigValue("ORCHESTRATOR_REDIRECT_URL", "http://localhost:8083/callback");
+    public static final String ORCHESTRATOR_JAR_SIGNING_JWK =
+            getConfigValue("ORCHESTRATOR_JAR_SIGNING_JWK", "missing-jar-signing-jwk");
     public static final String ORCHESTRATOR_CLIENT_SIGNING_KEY =
             getConfigValue("ORCHESTRATOR_CLIENT_SIGNING_KEY", "missing-key");
+    public static final String ORCHESTRATOR_CLIENT_ENCRYPTION_KEY =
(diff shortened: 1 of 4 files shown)


dannyhayes-gds (dismissed):

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
```diff
@@ -16,8 +16,12 @@ public class OrchestratorConfig {
             getConfigValue("ORCHESTRATOR_CLIENT_ID", "di-ipv-orchestrator-stub");
     public static final String ORCHESTRATOR_REDIRECT_URL =
             getConfigValue("ORCHESTRATOR_REDIRECT_URL", "http://localhost:8083/callback");
+    public static final String ORCHESTRATOR_JAR_SIGNING_KEY =
+            getConfigValue("ORCHESTRATOR_JAR_SIGNING_KEY", "missing-jar-signing-key");
```
dannyhayes-gds: In fact I think we should make a change here. We won't need another signing key. Ideally we should just use the one below on line 21 instead and just update that certificate to be the correct format

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JWTSigner.java
```diff
@@ -0,0 +1,102 @@
+package uk.gov.di.ipv.stub.orc.utils;
+
+import com.nimbusds.jose.*;
```
dannyhayes-gds: Also can we remove the star import here. I think this is just a setting in intellij that you need to change otherwise it will keep auto importing the whole package when you use over like 3 or 4 dependencies

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
dannyhayes-gds: We'd need to make a minor change somewhere else in the orc stub to make sure that we are using EC for all of our signing. But we can talk about that tomorrow

dannyhayes-gds (approved):

PR #99, opened 2022-05-18
PYIC-1048 Update VC to conform to RFC0024

blakeyp:
### What changed
Updates to conform to [rfc-0024](https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0024-identity-vc-for-checks.md) and based on similar [passport CRI changes](https://github.com/alphagov/di-ipv-cri-uk-passport-back/pull/127)
- renames of evidence scoring params
- addition of required `type` and `txn` properties in evidence object
- removal of `@context` properties no longer needed

### Why did it change

To conform to rfc-0024 and align with real CRIs

### Issue tracking
- [PYIC-1048](https://govukverify.atlassian.net/browse/PYIC-1048)


dannyhayes-gds:
The only change I can think of that we might need here is around contra indicators. Essentially if something goes wrong with the identity verification process we would receive a contra indicator. So maybe we need a check here if the user has entered a 0 in the gpg45 score then we add a CI value. 

If this is what we want then we can handle that with a separate jira ticket subtask and PR though as I want to try get this merged to fix our other vot bug. Other than this it looks all good though I think

dannyhayes-gds (approved):

PR #103, opened 2022-05-19
PYIC-1126: Add datasets

AmritSidhu:
### What changed

Added JSON file containing cri stub data of 3 different credentials

Used JQuery to add example datasets to a select list, when item selected from list it then populates the textarea with correct json relating to a specific credential issuer and credential

Each credential has been given a `criType`, this is used to match with the `cri-name`. With this it adds the stub data relating to the cri to the dropdown list
Screenshot 2022-05-19 at 15 01 28
Screenshot 2022-05-19 at 15 01 34

### Why did it change

This makes it more convenient when we want to test credentials

### Issue tracking

- [PYI-1126](https://govukverify.atlassian.net/browse/PYI-1126)

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/resources/templates/authorize.mustache
+++ b/di-ipv-credential-issuer-stub/src/main/resources/templates/authorize.mustache
@@ -25,6 +25,12 @@
 <script>
     document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
 </script>
+<script
+        src="https://code.jquery.com/jquery-3.6.0.min.js"
+        integrity="sha256-/xUj+3OJU5yExlq6GSYGSHk7tPXikynS7ogEvDej/m4="
+        crossorigin="anonymous">
+
+</script>
 <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
 
 <header class="govuk-header " role="banner" data-module="govuk-header">
@@ -65,6 +71,8 @@
                 </div>
             </details>
         {{/shared_claims}}
+
+        <div class="govuk-form-group" id="test_data_block"></div>
 
         {{#hasError}}
             <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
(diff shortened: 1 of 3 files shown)


di-ipv-credential-issuer-stub/src/main/resources/templates/authorize.mustache
```diff
@@ -208,5 +216,27 @@
...
+
+        var label = $('<label class="govuk-label" for="sort">').text("Select CRI stub data: ");
+        $('#test_data_block').append(label).append(select);
+        $("#test_data").on("change", function(event) {
+            var val = $(event.target).val();
+            var {payload} = filtered.find(x => x.label === val)
+            $("#jsonPayload").val(JSON.stringify(payload, undefined, 4))
+        });
+    });
+</script>
```
Wynndow, line 239: Would this be better as a separate JS file, rather than inlined like this?

di-ipv-credential-issuer-stub/src/main/resources/templates/authorize.mustache
```diff
@@ -66,6 +72,8 @@
             </details>
         {{/shared_claims}}
 
+        <div class="govuk-form-group" id="test_data_block"></div>
```
Wynndow, line 75: There's a possibility this will break the smoke tests as it may alter the x-paths to the input fields below. It depends how the smoke tests are finding the elements.
It might be worth running this up locally and checking the xpaths the smoke tests are using still find the right things.

Wynndow: Although looking at the tests, they're using id's rather than xpaths on that page so it might be ok...
https://github.com/alphagov/di-ipv-core-tests/blob/main/src/test/java/gov/di_ipv_core/pages/SupplyDataPage.java

Wynndow (approved):
Nice! Let's give it a whirl

di-ipv-credential-issuer-stub/src/main/resources/templates/authorize.mustache
Wynndow: From a slack chat "For the small amount of javascript it will be fine, otherwise we will have to deal with cache busting etc if we had a separate js file"

PR #105, opened 2022-05-19
PYIC-1125: Allow OAuth errors to be triggered in CRI stub

Wynndow:
### What changed

Allow OAuth errors to be triggered in CRI stub

### Why did it change

To allow us to more easily test journeys in the core where a CRI has returned an error.

One thing this piece of work doesn't do is allow the return of an
error response on the initial call to the auth endpoint. It's possible
for an error to happen, but we'd need a way to "prime" the CRI before
getting to the page presented by a GET to the auth endpoint. Some 
sort of backchannel request. Maybe this could be looked at in future.

The tests are fairly minimal and don't cover all use cases. The new
service is also being tested implicitly rather than directly. It's also
not being tested across an actual journey between the auth and token
endpoints. It would require a functional test with an app which is more
effort that it's probably worth right now.

### Issue tracking

- [PYI-1125](https://govukverify.atlassian.net/browse/PYI-1125)

#### Screenshot of the CRI stub with changes
Screenshot 2022-05-19 at 21 55 23

### DIFF
--- a/di-ipv-core-stub/startup.sh
+++ b/di-ipv-core-stub/startup.sh
@@ -15,4 +15,4 @@ fi
 cp -r $CONFIG_DIR config
 
 docker build -t ipv-core-stub .
-docker run -p 8085:8085 --env-file ${CONFIG_DIR}/.env ipv-core-stub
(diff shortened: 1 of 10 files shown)


Danomanic (approved):
LGTM

PR #115, opened 2022-06-10
KBV-484: Minor patch to allow testing of fraud check with users who have house names

gds-jbentham:
### What changed

Mapped house name to street 1 if house house number is not present

### Why did it change

To ease the testing of fraud

### DIFF
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
@@ -43,6 +43,7 @@ public class IdentityMapper {
         UKAddress address =
                 new UKAddress(
                         houseNo,
+                        map.get("houseName"),
                         map.get("street"),
                         map.get("district"),
                         map.get("posttown"),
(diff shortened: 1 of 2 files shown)


monhaque (dismissed):

di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
```diff
@@ -40,6 +40,10 @@ public Identity map(Map<String, String> map, int rowNumber) {
             houseNo = (String) hn;
         }
 
+        if (houseNo == null || houseNo.equals("")) {
```
tal-nagra: It would be better to modify UKAddress so that it looks like:
```
public record UKAddress(
        String buildingNumber,
        String buildingName,
        String street,
        String county,
        String townCity,
        String postCode,
        boolean currentAddress) {}
```
The existing `houseNo` can be mapped to buildingNumber

The houseName can be mapped to buildingName:

```
String houseName = null;
        if (StringUtils.isBlank(houseNo)) {
             houseName = map.get("houseName");
        }

        UKAddress address =
                new UKAddress(
                        houseNo,
                        houseName,
                        map.get("street"),
                        map.get("district"),
                        map.get("posttown"),
                        map.get("postcode"),
                        true);
```

Then in the `mapToSharedClaim` method:

```
new CanonicalAddress(
                                identity.UKAddress().buildingNumber(),
                                identity.UKAddress().buildingName(),
                                identity.UKAddress().street(),
                                identity.UKAddress().townCity(),
                                identity.UKAddress().postCode(),
                                // default / arbitrary value assigned for now as
                                // the validFrom date is not available in test data
                                LocalDate.of(2021, 1, 1),
                                null)));
```

tal-nagra (changes requested):

tal-nagra (approved):

PR #116, opened 2022-06-13
PYIC-167 Update to allow providing contraindicators via text field

blakeyp:
### What changed

Add field to input list of contraindicators via the CRI stub

### Why did it change

So that we can set contraindicators on the VC

### Issue tracking
- [PYIC-167](https://govukverify.atlassian.net/browse/PYIC-167)


Danomanic (approved):
Approved and seen working locally.

PR #141, opened 2022-08-04
LIME-38 - Stub update to simulate PEP checking.

MMcC-GDS:
### What changed

Added PEPs support to stub.
Added PEPs JSONs.
Updated Dockerfile and startup.sh.

### Issue tracking

https://govukverify.atlassian.net/browse/LIME-38

```
P1-36 

```


experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Config.java
```diff
@@ -1,26 +1,42 @@
...
+    static {
+        ciList = List.of("CI1", "CI2", "CI3", "CI4", "CI5");
+        ciMap = new HashMap<>();
+        for (String ci : ciList) {
+            ciMap.put(ci, getConfigValue(ci));
+        }
+
+        pepList =
+                List.of(
+                        "P1", "P2", "P3", "P4", "P5", "P6", "P7", "P8", "P9", "P10", "P11", "P12",
```
ElliotMurphyGDS: Makes sense for testing purposes that these reflect each of the possible 11 pep codes with 1-3 for each code...I wonder though if there is a neater way of doing this in the future 🤔

experian-fraud-stub/src/main/resources/GenericResponse/fraud-ex--peps.json
```diff
@@ -0,0 +1,295 @@
...
+        "applicantId": "APPLICANT_1",
+        "decision": "NA00",
+        "score": 20,
+        "decisionText": "",
+        "decisionReason": "The Applicant has not been Authenticated to your required minimum level. You should seek o",
+        "appReference": "6FXPN3G9WP",
+        "rules": [
+          {
+            "ruleId": "P1",
+            "ruleName": "PEPs",
```
ElliotMurphyGDS, line 48: Looks good just wondering if this ruleName is ok/will it need changing? We could change it to be the Pcode we are returning i.e P1 to be inline with the two specific Json responses but not a big issue as this won't be in production and we are only testing against ruleId

ElliotMurphyGDS (dismissed):

experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Config.java
MMcC-GDS: Agreed.

experian-fraud-stub/Dockerfile
```diff
@@ -11,6 +11,79 @@ ENV CI3=$CI3
 ENV CI4=$CI4
 ENV CI5=$CI5
 
+ARG P1
```
gds-jbentham: Could we pass this as a string or json blob and parse in a bash script or java

MMcC-GDS: Updated.

experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Config.java
MMcC-GDS: Updated.

gds-jbentham (approved):

PR #145, opened 2022-08-22
OJ-883 fix Experian Stub incorrect questions response data

idrop:
## Proposed changes

The Experian Stub asks two questions, unlike the real Experian API. This is ok, but we expect to issue a V03 CI when then there is more than one questions answered incorrectly.

Fix the Stub response for the “answered incorrectly” case so that the number of incorrectly answered questions is 2.

### Issue tracking

- [OJ-883](https://govukverify.atlassian.net/browse/OJ-883)

### DIFF
--- a/experian-kbv-stub/build.gradle
+++ b/experian-kbv-stub/build.gradle
@@ -10,7 +10,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


MattBidewell (approved):
Small and simple

PR #159, opened 2022-09-30
OJ-979: Backend only to add postcode claim

jkunle:
### What changed

Add a new routes to core-stub i.e. /backend/generateInitialClaimSetPostCode

Minor refactoring to allow backend/frontend journey to share the same code path.

### Why did it change
To enable starting a backend only journey for postcode claim

Tested for AddressCRI via -

Making a Get request to http://localhost:8085/backend/generateInitialClaimsSetPostCode?cri=address-cri-dev&postcode=
Which will return a complete ClaimSet (JSON) with the postcode claims populating address region with the postcode only

Post the JSON/ClaimsSet to http://localhost:8085/backend/createSessionRequest?cri=address-cri-dev

This will then create a JWT using the claimSet which is then signed and encrypted.
A JSON reply (AuthorizationRequest) will be returned in the format {"request":", "Data, client_id":"ipv-core-stub"}, which can then be posted to the session resource on the private api gateway.
Outputted in the logs will be the equivalent AuthorizationRequest URI which (if the cri front end is running) allow picking up the journey in the frontend for debugging.

For dev testing the following script was used for posting the JSON/ClaimsSet back.

- [OJ-979](https://govukverify.atlassian.net/browse/OJ-979)

### DIFF
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/CoreStub.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/CoreStub.java
@@ -57,7 +57,9 @@ public class CoreStub {
         Spark.get(
                 "/backend/createTokenRequestPrivateKeyJWT",
                 coreStubHandler.createTokenRequestPrivateKeyJWT);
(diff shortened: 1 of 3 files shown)


techaman-m (approved):
LGTM

PR #168, opened 2022-11-29
LIME-257 Fix surnames overriding response type and use a copy of the response template to stop influencing responses to later requests.

smsgds:
### What changed

Separated handling of fraud and pep requests.

Make a copy of the response template and then make changes to the copy.

Added test surname FSC_ and PSC_  which when appended e.g. FSC_500 sets the response status code. (status code if not set defaults to 200)

### Why did it change

Separated requests to avoid special test surnames interacting and returning the incorrect response type.

Changes where being made directly to the template response, which work as long as all fields where overwritten each time but means some data was latched when it wasn't.

FSC_ / PSC_ to allow testing API behaviour to various http status codes in both request separately.

### Issue tracking

- [LIME-257](https://govukverify.atlassian.net/browse/LIME-257)

### DIFF
--- a/experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Handler.java
+++ b/experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Handler.java
@@ -1,5 +1,6 @@
 package uk.gov.di.ipv.stub.fraud;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -16,11 +17,17 @@ public class Handler {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(Handler.class);
     private ObjectMapper mapper;
-    private InMemoryDataStore inMemoryDataStore;
+
+    public static final String FRAUD_CHECK_SOURCE = "fraud_check_source";
+    private InMemoryDataStore fraudCheckInMemoryDataStore;
+
+    public static final String PEP_CHECK_SOURCE = "pep_check_source";
+    private InMemoryDataStore pepCheckInMemoryDataStore;
 
     protected Handler() {
         mapper = new ObjectMapper();
-        inMemoryDataStore = new InMemoryDataStore(mapper);
+        fraudCheckInMemoryDataStore = new InMemoryDataStore(mapper, FRAUD_CHECK_SOURCE);
+        pepCheckInMemoryDataStore = new InMemoryDataStore(mapper, PEP_CHECK_SOURCE);
     }
 
     protected Route root = (Request request, Response response) -> "ok";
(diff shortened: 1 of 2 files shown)


experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/InMemoryDataStore.java
```diff
@@ -19,8 +19,8 @@ public class InMemoryDataStore {
     private static final Logger LOGGER = LoggerFactory.getLogger(InMemoryDataStore.class);
     private final HashMap<String, IdentityVerificationResponse> experianResponses = new HashMap<>();
 
-    public InMemoryDataStore(ObjectMapper mapper) {
-        init();
+    public InMemoryDataStore(ObjectMapper mapper, boolean forPep) {
```
gds-jbentham: Rather than a bool could we pass a string so this is a bit more extendable

experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Handler.java
```diff
@@ -37,67 +40,45 @@ protected Handler() {
...
                         .getRequestType()
                         .equals("Authenticateplus-Standalone")) {
+
+                    // Look for a response for specific user surname, fall back to AUTH1
+                    modifiableResponse =
+                            getModifiableResponse(
+                                    requestNames.get(0).getSurName().toUpperCase(), "AUTH1", false);
+
+                    // Decision score set to suffix
+                    if (requestNames.get(0).getSurName().contains("NO_FILE_")) {
```
gds-jbentham, line 63: Would all these ifs now be better as a case and in a private method

gds-jbentham: Actually ignore private method that would be worse mutation

experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/Handler.java
```diff
@@ -184,6 +200,27 @@ protected Handler() {
                 LOGGER.info("Fraud request ID: " + fraudResponseId);
 
                 response.header("Content-Type", "application/json");
-                return mapper.writeValueAsString(inMemoryDataStore.getResponse(fraudResponseId));
+                return mapper.writeValueAsString(
+                        fraudCheckInMemoryDataStore.getResponse(fraudResponseId));
             };
+
+    private IdentityVerificationResponse getModifiableResponse(
```
gds-jbentham, line 213: Same as datastore bool use

experian-fraud-stub/src/main/java/uk/gov/di/ipv/stub/fraud/InMemoryDataStore.java
smsgds: Changed to predefined strings - used here and in the check lookup.

ElliotMurphyGDS (approved):

gds-jbentham (approved):

PR #171, opened 2022-12-19
BAU Add input field for custom evidence block

blakeyp:
### What changed

Add input field for custom evidence payload in the CRI stubs (every CRI except address) and allow choosing from a dropdown
Screenshot 2022-12-21 at 12 16 18

### Why did it change

Allows us to override the evidence params with a custom json evidence payload so we can test with new fields like checkDetails

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
+++ b/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
@@ -1,0 +1,87 @@
+{
+  "data": [
+    {
+      "criType": "Fraud Check (Stub)",
+      "label": "Passed fraud check (M1A)",
+      "payload": {
+        "type": "IdentityCheck",
+        "identityFraudScore": 2
+      }
+    },
+    {
+      "criType": "UK Passport (Stub)",
+      "label": "Passed passport check (M1A)",
+      "payload": {
+        "type": "IdentityCheck",
+        "validityScore": 2,
+        "strengthScore": 4
+      }
+    },
+    {
+      "criType": "Knowledge Based Verification (Stub)",
+      "label": "Passed verification (M1A) with checkDetails",
+      "payload": {
+        "type": "IdentityCheck",
+        "verificationScore": 2,
+        "checkDetails": [
+          {
+            "kbvResponseMode": "multiple_choice",
+            "kbvQuality": 3,
+            "checkMethod": "kbv"
+          },
+          {
+            "kbvResponseMode": "multiple_choice",
+            "kbvQuality": 3,
+            "checkMethod": "kbv"
+          },
+          {
+            "kbvResponseMode": "multiple_choice",
+            "kbvQuality": 2,
+            "checkMethod": "kbv"
+          }
+        ]
+      }
+    },
+    {
+      "criType": "Knowledge Based Verification (Stub)",
+      "label": "Failed verification with failedCheckDetails",
+      "payload": {
+        "type": "IdentityCheck",
+        "verificationScore": 0,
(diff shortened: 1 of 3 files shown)


kerrr (dismissed):

galund:
For what it's worth, I'd prefer there to be a set of labelled evidence blocks in config, from which you could choose one. That way, pre-determined scenarios can more easily be tested?

di-ipv-credential-issuer-stub/src/main/resources/templates/authorize.mustache
```diff
@@ -368,6 +380,26 @@
             $("#jsonPayload").val(JSON.stringify(payload, undefined, 4))
         });
     });
+
+    var evidenceDropdownData = {{{cri_stub_evidence_payloads}}}
```
galund, line 384: this isn't end-user facing code, but if it was, I'd be moaning about a needless (AFAICS) dependency on JavaScript to build the contents of a dropdown ;-)

di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
```diff
@@ -0,0 +1,87 @@
...
+          {
+            "kbvResponseMode": "multiple_choice",
+            "checkMethod": "kbv"
+          }
+        ]
+      }
+    },
+    {
+      "criType": "DOC Checking App (Stub)",
+      "label": "Passed DCMAW (M1B) with checkDetails",
```
galund: There are examples of failed checks for DCMAW in Confluence:

https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3246948417/DCMAW+-+Verifiable+Credential+Response+-+Passport#Example-2---Unsuccessful-NFC-chipped-passport-biometric-verification

https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3141435393/DCMAW+-+Verifiable+Credential+Response+-+Driving+License#Example-4---failed-Document-Check-biometric-verification

https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3141435393/DCMAW+-+Verifiable+Credential+Response+-+Driving+License#Example-5---Failed-JWT---Document-Valid---Liveness%2FLikeness-Failed---non-Fraud

galund (commented):
Thanks, this is exactly what I had in mind! 🙏

di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
blakeyp: Thanks I've added those

di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
```diff
@@ -0,0 +1,147 @@
...
+        "failedCheckDetails": [
+          {
+            "checkMethod": "vri",
+            "identityCheckPolicy": "published"
+          },
+          {
+            "checkMethod": "bvr",
+            "biometricVerificationProcessLevel": 3
+          }
+        ]
```
galund, line 130: I think you want to include the contra-indicators here otherwise it's basically the same as the next example. The distinction is supposed to be between "fraud and non-fraud" cases, which I think is a bit dubious (we're likely to come back to the 'non-fraud' version at some point as it's a bit meaningless - a contra-indicator does not necessarily have to be due to fraud).

blakeyp: Good point, I've added the CIs now

kerrr (approved):

PR #309, opened 2023-03-22
Pyic 2353 migrate di ipv orchestrator stub to aws stubs account create the required deployment

TobySaundersGDS:
Updated dns config to remove space,, set ECS lambda to use programmatic permissions boundary


TalatAydinGDS (approved):
Looks good to me.

PR #310, opened 2023-03-22
PYIC-2583 Add stub DL VC fail example

blakeyp:
Add stub driving licence VC fail example

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
+++ b/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
@@ -168,6 +168,23 @@
           }
         ]
       }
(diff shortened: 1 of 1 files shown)


di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
```diff
@@ -168,6 +168,23 @@
...
+        "failedCheckDetails": [
+          {
+            "identityCheckPolicy": "published",
+            "checkMethod": "data"
+          }
+        ]
+      }
     }
   ]
 }
```
kerrr, line 190: For the future can we make sure we have newlines at the end of files.

kerrr (approved):

PR #397, opened 2023-07-03
BAU Check in package-lock to fix queue stub build

blakeyp:
BAU Check in package-lock to fix queue stub build

### DIFF
--- a/di-ipv-queue-stub/enqueue-event/package.json
+++ b/di-ipv-queue-stub/enqueue-event/package.json
@@ -7,9 +7,7 @@
         "@aws-sdk/client-secrets-manager": "^3.357.0",
         "@aws-sdk/client-sqs": "^3.357.0",
         "@aws-sdk/client-ssm": "^3.357.0",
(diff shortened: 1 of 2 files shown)


di-ipv-queue-stub/enqueue-event/package.json
```diff
@@ -7,9 +7,7 @@
         "@aws-sdk/client-secrets-manager": "^3.357.0",
         "@aws-sdk/client-sqs": "^3.357.0",
         "@aws-sdk/client-ssm": "^3.357.0",
-        "@types/dotenv": "^8.2.0",
-        "@types/node": "^20.2.5",
-        "@types/uuid": "^9.0.2"
```
Wynndow, line 12: Do we not need `dotenv` or `uuid` anymore?

blakeyp: They're not being used - think they were just copied over from another package.json

Wynndow (approved):

PR #440, opened 2023-08-02
Pyic 3295 secure the network access to the stub

TobySaundersGDS:
This adds a RESTAPI gateway to the CIMIT stub with a usage plan, and an API key

### What changed
Added an API key as a basic barrier to entry

### Why did it change
To prevent accidental discovery of the CIMIT Stub service 

### Other considerations
The API definition needs fleshing out - currently just defining /healthcheck.
If this needs to be called within the VPC we will need to define DNS for it as otherwise traffic is blackholed.


di-ipv-cimit-stub/deploy/template.yaml
```diff
@@ -317,3 +339,50 @@ Resources:
...
+          Name: "AWS::Include"
+          Parameters:
+            Location: "../openAPI/cimit-external.yaml"
+      ApiKeySourceType: HEADER #pragma: allowlist secret
+      Auth:
+        ApiKeyRequired: true #pragma: allowlist secret
+        UsagePlan:
+          CreateUsagePlan: PER_API
+          UsagePlanName: "CIMIT Usage Plan"
+        #ResourcePolicy:
```
therealslimsladey: Remove commented out config?

.cfnlintrc.yaml
```diff
@@ -0,0 +1,2 @@
+ignore_checks:
+  - W3005
```
therealslimsladey, line 2: Can this be a lint config for cimit stub only?

therealslimsladey (dismissed):
lgtm, minor comments only

.cfnlintrc.yaml
TobySaundersGDS: I decided to leave it as I expect to be doing a similar thing for the orch stub and maybe others (also just moving it to the template folders  didn't work)

di-ipv-cimit-stub/deploy/template.yaml
TobySaundersGDS: yes

lwebb-iig (approved):

PR #444, opened 2023-08-02
PYIC-3058: add error payload for f2f stub

shivanshuit914:
## Proposed changes

- Added option to send error message to async queue

### Issue tracking

- [PYI-3058](https://govukverify.atlassian.net/browse/PYI-3058)

### Other considerations

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
+++ b/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
@@ -339,7 +339,8 @@ public class AuthorizeHandler {
                                     "checked");
                     boolean F2F_SEND_ERROR_QUEUE =
                             Objects.equals(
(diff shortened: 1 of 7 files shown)


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/handlers/AuthorizeHandler.java
```diff
@@ -357,6 +360,26 @@ public AuthorizeHandler(
                         HTTPResponse httpResponse = httpRequest.send();
                     }
 
+                    if (F2F_SEND_VC_QUEUE && F2F_SEND_ERROR_QUEUE) {
```
blakeyp: Thinking should it just be if `F2F_SEND_ERROR_QUEUE` checkbox is checked not both?

shivanshuit914: hm I have added option to make sure we select queue and error queue. but yes makes sense we can get rid of the option. thanks

blakeyp (approved):

PR #479, opened 2023-09-04
BAU: Make CRI stub health check less noisy

Wynndow:
### What changed

Make CRI stub health check less noisy

### Why did it change

The ALB healthcheck for the CRI is configured to send a GET request to `/` and check for a 404 http status response.

This route isn't mapped in Spark which generates the required 404, but it does fill the logs with junk moaning about the route not being mapped.

This adds a handler for that route that returns a 200 response, and updates the healthcheck to expect a 200 instead of a 404.


kerrr (approved):
🚀

PR #486, opened 2023-09-07
BAU: Provide stubs-prod CiMit stub URL as default

Wynndow:
### What changed

Provide stubs-prod CiMit stub URL as default

### Why did it change

The dev and build envs are going to be configured to use the stubs-prod deploy of out CiMit stub. It would be helpful to have the correct URL for the mitigations admin endpoint already configured. Think of the cost saving for all those less worn keyboard keys...


therealslimsladey (approved):
lgtm

PR #513, opened 2023-10-05
Pyic 3321 orch stub basic auth

DanCorderIPV:
### What changed

By default any newly deployed orchestrator stub will require an SSM parameter specifying a username and password for basic auth, and any requests to the stub will need to provide the correct username and password to view the stub pages.

With this PR all dev orch stubs will default to credentials of dev/dev which can be overridden in core-common-infra. This could be changed to force developers to set up SSM credentials manually instead depending on how strictly we want to enforce this.

### Why did it change

Having all stubs orch publicly accessible isn't a good thing.

### Issue tracking

- [PYI-3321](https://govukverify.atlassian.net/browse/PYI-3321)

For this code to work every non-dev orch stub will need an SSM parameter configured. The exact parameter path will depend on the environment but takes the form of:

`/stubs/<environment>/orch/env/ORCHESTRATOR_BASIC_AUTH_CREDENTIALS`

an example value is:

`{ "username": "dev", "password": "dev" }`

These values are secret and so need to be added manually to SSM before this PR is merged.

- [x] Documented in the [README](./blob/main/README.md)

### Other considerations

- [x] Update [README](./blob/main/README.md) with any new instructions or tasks

### DIFF
--- a/di-ipv-orchestrator-stub/deploy-core-dev/template.yaml
+++ b/di-ipv-orchestrator-stub/deploy-core-dev/template.yaml
@@ -48,6 +48,18 @@ Parameters:
     Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
     Type: AWS::SSM::Parameter::Value<String>
     Default: "/stubs/build/orch/env/ORCHESTRATOR_JAR_ENCRYPTION_PUBLIC_KEY" #pragma: allowlist secret
+  OrchestratorBasicAuthEnable:
+    Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: "/stubs/build/orch/env/ORCHESTRATOR_BASIC_AUTH_ENABLE" #pragma: allowlist secret
+  OrchestratorBasicAuthUsername:
+    Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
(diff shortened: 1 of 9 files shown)


di-ipv-orchestrator-stub/deploy-core-dev/template.yaml
```diff
@@ -76,6 +76,14 @@ Parameters:
...
     Type: AWS::SSM::Parameter::Value<String>
     Default: "/stubs/build/orch/env/JBP_CONFIG_OPEN_JDK_JRE" #pragma: allowlist secret
+  OrchestratorEnableBasicAuth:
+    Description: "Environment Variables for ECS"
+    Type: String
+    Default: "true" #pragma: allowlist secret
+  OrchestratorBasicAuthCredentials:
+    Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
+    Type: String
+    Default: "{ \"username\": \"dev\", \"password\": \"dev\" }" #pragma: allowlist secret
```
DanCorderIPV: If this is too insecure we can change it match the other templates. This would mean that developers would need to add the relevant SSM parameter before they could re-deploy their orch stubs.

DanCorderIPV:
Requesting reviews from Patrick as this is to do with security, and Vam as it adds config and he's worked in this area recently.

blakeyp (dismissed):

blakeyp (approved):

PR #528, opened 2023-10-18
PYIC-3740: Add NINO CRI payloads for identity checking

DanCorderIPV:
### What changed

Extra evidence blocks for the NINO CRI along with renaming of existing ones so that it's clear when to use them.

### Why did it change

NINO CRI will be changing behaviour soon, this should allow us to stub the new behaviour

### Issue tracking

- [PYI-3740](https://govukverify.atlassian.net/browse/PYI-3740)

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
+++ b/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
@@ -353,7 +353,7 @@
     },
     {
       "criType": "National Insurance Number (Stub)",
-      "label": "Passed National Insurance Number (Stub)",
+      "label": "Passed National Insurance Number - no scope",
       "payload": {
         "type": "IdentityCheck",
         "checkDetails": [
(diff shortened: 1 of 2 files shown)


di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
```diff
@@ -374,6 +374,49 @@
...
+      }
+    },
+    {
+      "criType": "National Insurance Number (Stub)",
+      "label": "Failed National Insurance Number with CI - 'identityCheck' scope",
+      "payload": {
+        "type": "IdentityCheck",
+        "strengthScore": 2,
+        "validityScore": 0,
+        "ci": ["QQQ"],
```
DanCorderIPV, line 413: This will need to be changed once the possible CIs for NINO are agreed.

DanCorderIPV:
The NINO CRI Stub currently doesn't allow for individual entry of strength and validity numbers.
I decided to leave that alone as it seems like selecting a complete valid evidence block (and tweaking if necessary) is less error-prone. However that can be changed if necessary.

shivanshuit914 (approved):

PR #544, opened 2023-10-24
PYIC-3781: Update cimit stub mitigation handling

Wynndow:
### What changed

Updates the way the cimit stub handles mitigations. Any new mitigation requests
are stored as pending requests in a table. The mitigation is only applied once the
post mitigation endpoint receives the expected VC. This is determined by the VCs
JTI which will have been sent in the initial request from the CRI stub.

### Why did it change

To more accurately reflect how we handle mitigations in production and to avoid
weird race conditions.

### Issue tracking

- [PYI-3781](https://govukverify.atlassian.net/browse/PYI-3781)


di-ipv-cimit-stub/lambdas/stub-management/src/main/java/uk/gov/di/ipv/core/stubmanagement/StubManagementHandler.java
```diff
@@ -73,10 +87,12 @@ public APIGatewayProxyResponseEvent handleRequest(
                 String ci = pathParameters.get(CI_PATH_PARAMS);
                 UserMitigationRequest userMitigationRequest =
                         objectMapper.readValue(event.getBody(), UserMitigationRequest.class);
-                if (httpMethod.equals(HttpMethod.POST.toString())) {
-                    userService.addUserMitigation(userId, ci, userMitigationRequest);
-                } else if (httpMethod.equals(HttpMethod.PUT.toString())) {
-                    userService.updateUserMitigation(userId, ci, userMitigationRequest);
+                if (SUPPORTED_MITIGATION_METHODS.contains(httpMethod)) {
```
Joe-Edwards-GDS, line 90: Do we actually use the `PUT` version for anything? Feels a bit odd to overwrite a previous mitigation, especially now that mitigations occur asynchronously when you receive the corresponding VC.

E.g. if you did:
- PUT mitigation 01
- POST mitigation 02
- Receive VC for mitigation 02
- Receive VC for mitigation 01

You'd end up with only mitigation 01 - which might be surprising!

Joe-Edwards-GDS (approved):
Looks good, with one comment about whether we really need the 'PUT' version of the mitigation.

Also worth noting: this will only apply mitigations to _existing_ CIs, it will not mitigate if the CI is received after the mitigation. I don't think this is possible with the journey config, but perhaps will be in future - and we can worry about it then.

di-ipv-cimit-stub/lambdas/stub-management/src/main/java/uk/gov/di/ipv/core/stubmanagement/StubManagementHandler.java
Wynndow: I think the initial idea might have been to allow QA's to easily get stuff set up how they want it. I think it might have been one of those things that seems like a good idea, but in reality doesn't get used. 
We'd never use the PUT in a journey flow - it's more about manually tweaking stuff.

PR #569, opened 2023-11-15
PYIC-3978: Update the Orchestration Stub to include 'vtr' in authorisation request

TalatAydinGDS:
### What changed

Update the Orchestration Stub to include 'vtr' in authorisation request

### Why did it change

The "vtr" is a crucial assertion of trust expectations set by the Relying Party. Its secure storage within IPV Core is imperative for confirming that the identity proofing journey meets the RP's required trust benchmarks. This ensures that IPV Core can validate and relay trust expectations to the TICF CRI, maintaining the integrity of the identity verification process.

The asynchronous nature of the F2F journey means that the original vtr defined when initiating the F2F journey, and the associated VC’s stored should match those in the re-use journey once the user has visited the post office. 

### Issue tracking

- [PYIC-3978](https://govukverify.atlassian.net/browse/PYIC-3978)

### Other considerations

[PYIC-3978]: https://govukverify.atlassian.net/browse/PYIC-3978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JwtBuilder.java
+++ b/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JwtBuilder.java
@@ -30,6 +30,7 @@ import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Base64;
 import java.util.Date;
+import java.util.List;
 import java.util.UUID;
 
 import static uk.gov.di.ipv.stub.orc.config.OrchestratorConfig.IPV_CORE_AUDIENCE;
@@ -72,6 +73,7 @@ public class JwtBuilder {
                 .claim("govuk_signin_journey_id", signInJourneyId)
                 .claim("persistent_session_id", UUID.randomUUID().toString())
                 .claim("email_address", "dev-platform-testing@digital.cabinet-office.gov.uk")
+                .claim("vtr", List.of("Cl.Cm.P2", "Cl.Cm.PCL200"))
                 .build();
     }
 


rpayal (approved):

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JwtBuilder.java
```diff
@@ -72,6 +73,7 @@ public static JWTClaimsSet buildAuthorizationRequestClaims(
                 .claim("govuk_signin_journey_id", signInJourneyId)
                 .claim("persistent_session_id", UUID.randomUUID().toString())
                 .claim("email_address", "dev-platform-testing@digital.cabinet-office.gov.uk")
+                .claim("vtr", List.of("Cl.Cm.P2", "Cl.Cm.PCL200"))
```
Joe-Edwards-GDS, line 76: The CL200 part corresponds to a specific HMRC inherited identity - I think the default value should be `["Cl.Cm.P2"]`, with an option to override it with something else.

(That option could also be left until a later ticket, since we won't need it until inherited identities comes along)

TalatAydinGDS: Thanks Joe for your comment and informations and raised a new PR what The 'vtr' value is now being retrieved with a default value on the front end. It can be changed if desired. https://github.com/govuk-one-login/ipv-stubs/pull/570

PR #572, opened 2023-11-16
PYIC-3978: fixed to the error

TalatAydinGDS:
### What changed

Update the Orchestration Stub to include 'vtr' in authorisation request

### Why did it change

fixed error 

### Issue tracking

- [PYIC-3978](https://govukverify.atlassian.net/browse/PYIC-3978)

### Other considerations

[PYIC-3978]: https://govukverify.atlassian.net/browse/PYIC-3978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
+++ b/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
@@ -33,7 +33,6 @@ import com.nimbusds.openid.connect.sdk.UserInfoResponse;
 import net.minidev.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import spark.QueryParamsMap;
 import spark.Request;
 import spark.Response;
 import spark.Route;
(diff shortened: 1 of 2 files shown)


di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -77,7 +78,13 @@ public class IpvHandler {
...
                 String userIdTextValue = request.queryMap().get("userIdText").value();
                 String signInJourneyIdText = request.queryMap().get("signInJourneyIdText").value();
-                String[] vtr = request.queryMap().get("vtrText").values();
+                QueryParamsMap queryParamsMap = request.queryMap().get("vtrText");
+                String[] vtr;
+                if (queryParamsMap != null) {
+                    vtr = queryParamsMap.values();
+                } else {
+                    vtr = new String[] {};
+                }
```
Joe-Edwards-GDS: I wonder if it'd be neater to use a ternary here, i.e.:
```
String[] vtr = request.queryMap().hasKey("vtrText")
        ? request.queryMap().get("vtrText").values()
        : new String[] {};
```

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -77,7 +78,13 @@ public class IpvHandler {
                 String errorType = request.queryMap().get("error").value();
                 String userIdTextValue = request.queryMap().get("userIdText").value();
                 String signInJourneyIdText = request.queryMap().get("signInJourneyIdText").value();
-                String[] vtr = request.queryMap().get("vtrText").values();
+                QueryParamsMap queryParamsMap = request.queryMap().get("vtrText");
+                String[] vtr;
+                if (queryParamsMap != null) {
+                    vtr = queryParamsMap.values();
+                } else {
+                    vtr = new String[] {};
```
Joe-Edwards-GDS: The problem Chris highlighted is that when the 'error' journey form is submitted, there is no VTR.

Your change will make it so that this case will return a default of empty VTR, but I think this isn't very representative. I wonder if we should make the default value `Cl.Cm.P2` as with the main form.

I think you could either:
- Set the default here, if the `vtrText` is not specified
- Add a hidden input to the error form, with the default

Joe-Edwards-GDS (changes requested):

Joe-Edwards-GDS (approved):

PR #575, opened 2023-11-20
PYIC-3894 Initial code changes for TiCF CRI stub to return TiCF VC. Added CF config files for dev and build env. And git workflow pipeline.

rpayal:
## Proposed changes
Creating a Stub of the TICF CRI

### What changed
- Created TICF stub
- Added ticf stub pipeline changes
- Added CF config for infrastructure creation and deployment of ticf stub

### Issue tracking

- [PYI-3894](https://govukverify.atlassian.net/browse/PYI-3894)

### Other considerations


rpayal:
@vam-gds pls. review the pipeline config for the new stub along with CF deployment config.

di-ipv-ticf-stub/README.md
```diff
@@ -0,0 +1,84 @@
+### TiCF CRI Stub
```
Joe-Edwards-GDS: I've not seen it stylised like this before - it's always just been 'TICF' to me.

di-ipv-ticf-stub/README.md
```diff
@@ -0,0 +1,84 @@
...
+This will set up an API gateway in front of a lambda and return TiCF cri VC.
+The format of the POST request to the API gateway should look like
```
Joe-Edwards-GDS, line 4: It might be worth linking to the confluence page here as well, for background.

di-ipv-ticf-stub/core-dev-deploy/template.yaml
```diff
@@ -0,0 +1,294 @@
...
+        Sourcemap: true # Enabling source maps will create the required NODE_OPTIONS environment variables on your lambda function during sam build
+        EntryPoints:
+          - src/handlers/ticfHandler.ts
+
+  GetTicfVCFunctionCoreDev01InvokePermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !Ref GetTicfVCFunction
+      Action: "lambda:InvokeFunction"
+      Principal: !FindInMap [ CoreAccounts, dev01, accountId ]
```
Joe-Edwards-GDS, line 125: Should there be one of these per environment?

di-ipv-ticf-stub/lambdas/src/common/ssmParameter.ts
```diff
@@ -0,0 +1,15 @@
...
+export const getSsmParameter = async (
+  name: string
+): Promise<string | undefined> => {
+  console.info(">> Getting value for ssm parameter: " + name);
+  let parameter;
+  try {
+    // Retrieve a single parameter
+    parameter = await getParameter(name);
+  } catch (error) {
+    console.error(`>> Error retriving parameter. Error: ${error}`);
```
Joe-Edwards-GDS: This means we'll just return `undefined` if fetching the parameter fails, which seems a bit surprising - is a missing parameter ever _not_ an error?

Perhaps we should be checking for undefined here instead, e.g.:
```
export const getSsmParameter = async (name: string): Promise<string> => {
  console.info(`Getting value for ssm parameter: ${name}`);
  // Retrieve a single parameter
  const parameter = await getParameter(name);
  if (parameter === undefined) {
    throw new Error(`Could not retrieve parameter: ${name}`);
  }
  return parameter;
};
```

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
```diff
@@ -0,0 +1,28 @@
+export const apiResponses = {
+  _200: (body: { [key: string]: any }) => {
+    return {
+      statusCode: 200,
+      body: JSON.stringify(body, null, 2),
+    };
+  },
+  _400: (body?: { [key: string]: any }) => {
+    if (!body) {
+      [(body = { errorMessage: "Bad request." })];
```
Joe-Edwards-GDS: I don't think you need the extra brackets here - could just be `body = { errorMessage: "Bad request." }`?

You could also inline this:
```
  _400: (body?: { [key: string]: any }) => ({
    statusCode: 400,
    body: JSON.stringify(body ?? { errorMessage: "Bad request" }, undefined, 2),
  }),
```

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
```diff
@@ -0,0 +1,28 @@
+export const apiResponses = {
+  _200: (body: { [key: string]: any }) => {
+    return {
+      statusCode: 200,
+      body: JSON.stringify(body, null, 2),
+    };
+  },
+  _400: (body?: { [key: string]: any }) => {
```
Joe-Edwards-GDS: This could probably just be `object`?

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+
+import { JWTPayload } from "jose";
+import { v4 as uuid } from "uuid";
+
+import { getSsmParameter } from "../common/ssmParameter";
+import { apiResponses } from "../common/apiResponses";
+import TicfRequest from "../domain/ticfRequest";
+import TicfResponse from "../domain/ticfResponse";
+
+export const handler: Handler = async (event: any) => {
```
Joe-Edwards-GDS: If you've specified the type of the function (i.e. `Handler`), you don't need to specify the type of the event explicitly.

In our case, we can actually use a more specific handler type, to give us type-safety over the requests and responses: `APIGatewayProxyHandlerV2`.

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+  try {
+    ticfRequest = JSON.parse(event.body);
+  } catch (error) {
+    return apiResponses._400();
+  }
+  if (ticfRequest == null)
+    return apiResponses._400({ errorMessage: "Pls. pass proper request." });
+
+  // preparing response
+  let ticfSigningKey: string | null | undefined = await getSsmParameter(
```
Joe-Edwards-GDS: In general we should use `const` for all variables, unless we need them to be mutable (but should try to avoid that)

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+  );
+  if (!ticfSigningKey) {
+    return apiResponses._500({
+      errorMessage: "Error while retrieving TicF CRI VC signing key.",
+    });
+  }
+  let timeoutVC: string | null | undefined = await getSsmParameter(
+    process.env.TICF_PARAM_BASE_PATH + "timeoutVC"
+  );
+  timeoutVC ??= "false";
```
Joe-Edwards-GDS: If you want to apply a default, you can do it in one go. However, I would question whether we actually want a default here - do we ever expect the SSM parameters not to be set?

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+  // preparing response
+  let ticfSigningKey: string | null | undefined = await getSsmParameter(
+    process.env.TICF_PARAM_BASE_PATH + "signingKey"
+  );
+  if (!ticfSigningKey) {
+    return apiResponses._500({
+      errorMessage: "Error while retrieving TicF CRI VC signing key.",
+    });
+  }
+  let timeoutVC: string | null | undefined = await getSsmParameter(
```
Joe-Edwards-GDS: We shouldn't need the types here - TS should infer it for you :)

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+  let buildJwtParams: SignedJwtParams;
+  buildJwtParams = {
+    issuer: process.env.ISSUER,
+    customClaims: getCustomClaims(
+      JSON.parse(timeoutVC.toLowerCase()),
+      JSON.parse(includeCIToVC.toLowerCase()),
+      ticfRequest.sub
+    ),
+    privateSigningKey: ticfSigningKey!,
+  };
```
Joe-Edwards-GDS: Similarly, why do we need to declare this separately to assignment? (ditto below)

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+          evidence: [
+            {
+              type: "RiskAssessment",
+              ci: ["V03"],
+            },
+          ],
+          type: ["VerifiableCredential", "RiskAssessmentCredential"],
+        },
+        jti: "urn:uuid:" + uuid(),
+      };
```
Joe-Edwards-GDS: We could probably pull out the evidence part, since the wrapper JWT is always the same

di-ipv-ticf-stub/lambdas/tsconfig.json
Joe-Edwards-GDS, line 1: Did you copy this from somewhere? I think we can probably simplify this a little - e.g. just including `src`, rather than excluding a bunch of stuff.

di-ipv-ticf-stub/openAPI/ticf-external.yaml
```diff
@@ -0,0 +1,91 @@
...
+        type: "MOCK"
+        requestTemplates:
+          application/json: "{\"statusCode\":200}"
+        responses:
+          200:
+            statusCode: 200
+            responseTemplates:
+              application/json: "{\"healthcheck\": \"ok\"}"
+
+  /:
```
Joe-Edwards-GDS: I'd be keen to give this a path if possible, just so it's easier to disambiguate from any other endpoints in future - perhaps `/risk-assessment`? (I don't know what the real one will look like yet)

di-ipv-ticf-stub/openAPI/ticf-external.yaml
```diff
@@ -0,0 +1,91 @@
...
+        type: "aws_proxy"
+        httpMethod: "POST"
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${GetTicfVCFunction.Arn}:live/invocations
+        passthroughBehavior: "when_no_match"
+        requestTemplates:
+          application/x-www-form-urlencoded:
+            Fn::Sub: |
+              {
+                "input": “{ \"vtr\": \"$input.body('vtr')\", \"vot\": \"$input.body('vot')\", \"vtm\": \"$input.body('vtm')\", \"sub\": \"$input.body('sub')\", \"govuk_signin_journey_id\": \"$input.body('govuk_signin_journey_id')\", \"https://vocab.account.gov.uk/v1/credentialJWT\": \"$input.body('https://vocab.account.gov.uk/v1/credentialJWT')\ "}"
```
Joe-Edwards-GDS: Do we need to map the request, can we just pass the body through as-is?

di-ipv-ticf-stub/lambdas/.eslintrc.js
Joe-Edwards-GDS, line 1: Does this work?

I think you'll need to use the `@typescript-eslint` plugin to handle typescript properly. You could look at the `delete-user-data` lambda in core-back as an example.

di-ipv-ticf-stub/lambdas/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "@types/aws-lambda": "^8.10.109",
+    "aws-sdk": "^2.1497.0",
+    "di-stub-oauth-client": "git+ssh://git@github.com:alphagov/di-stub-oauth-client#main",
+    "jose": "4.11.1",
+    "nodemon": "^3.0.1",
+    "ts-node": "^10.9.1",
+    "typescript": "^5.2.2"
+  },
+  "devDependencies": {
+    "eslint": "^8.54.0"
```
Joe-Edwards-GDS: Can we add a script to run eslint?

di-ipv-ticf-stub/lambdas/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "license": "GDS",
+  "dependencies": {
+    "@aws-lambda-powertools/parameters": "^1.16.0",
+    "@aws-sdk/client-ssm": "^3.451.0",
+    "@types/aws-lambda": "^8.10.109",
+    "aws-sdk": "^2.1497.0",
+    "di-stub-oauth-client": "git+ssh://git@github.com:alphagov/di-stub-oauth-client#main",
+    "jose": "4.11.1",
+    "nodemon": "^3.0.1",
+    "ts-node": "^10.9.1",
```
Joe-Edwards-GDS: Are we using `nodemon` or `ts-node`?

di-ipv-ticf-stub/lambdas/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "author": "IPVCore",
+  "license": "GDS",
+  "dependencies": {
+    "@aws-lambda-powertools/parameters": "^1.16.0",
+    "@aws-sdk/client-ssm": "^3.451.0",
+    "@types/aws-lambda": "^8.10.109",
+    "aws-sdk": "^2.1497.0",
+    "di-stub-oauth-client": "git+ssh://git@github.com:alphagov/di-stub-oauth-client#main",
```
Joe-Edwards-GDS: Has this repo been moved to `govuk-one-login`?

di-ipv-ticf-stub/core-dev-deploy/template.yaml
Joe-Edwards-GDS, line 1: I think it'd be good to get a review from one of the SREs on this (cc @vam-gds ), as there's a lot of stuff here and I'm not super close to our standards for it.

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+  return apiResponses._200(responseBody);
+};
+
+function getCustomClaims(
+  timeOutVC: boolean,
+  includeCIToVC: boolean,
+  userId: string
+): JWTPayload {
+  if (timeOutVC) {
+    if (includeCIToVC) {
```
Joe-Edwards-GDS: These two conditions are really mutually-exclusive - if the TICF API times out, then it cannot include any CIs.

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
Joe-Edwards-GDS: As a general rule for lambda handlers, I'd like to separate logic to parse the request and serialize the response from business logic.

In this case, I think the handler should do validation for the incoming request and construction of the `ApiGatewayProxyResult`, but all the logic around generating the signed JWTs should live in a service module.

Joe-Edwards-GDS (commented):
I think this is a good start, but I'm keen to set some solid patterns for TS lambdas, so I've left quite a few comments!

I think it's fine to use the SSM parameter approach as a way to test the core stub logic, but we need to make it dynamically configurable through a management API too.

Do you think we need to raise a separate sub-task for that change? Or we could even split out into a separate task to try and unblock the main TICF lambda development.

di-ipv-ticf-stub/README.md
rpayal: done

di-ipv-ticf-stub/README.md
rpayal: done

di-ipv-ticf-stub/core-dev-deploy/template.yaml
rpayal: done

di-ipv-ticf-stub/lambdas/src/common/ssmParameter.ts
rpayal: done

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
rpayal: done

di-ipv-ticf-stub/lambdas/package.json
rpayal: removed nodemon.....ts-node is needed to access environment variable.

di-ipv-ticf-stub/lambdas/package.json
rpayal: will check.

di-ipv-ticf-stub/core-dev-deploy/template.yaml
rpayal: yeah already added @vam-gds for review of pipeline and CF deployment config

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
rpayal: done

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
rpayal: done

di-ipv-ticf-stub/lambdas/package.json
rpayal: moved....so now using govuk-one-login.

rpayal:
> I think this is a good start, but I'm keen to set some solid patterns for TS lambdas, so I've left quite a few comments!
> 
> I think it's fine to use the SSM parameter approach as a way to test the core stub logic, but we need to make it dynamically configurable through a management API too.
> 
> Do you think we need to raise a separate sub-task for that change? Or we could even split out into a separate task to try and unblock the main TICF lambda development.

Yeah will be good to handle management api approach through a separate ticket. Do we need by storing value for timeout etc as a db value ?

Joe-Edwards-GDS:
> Yeah will be good to handle management api approach through a separate ticket. Do we need by storing value for timeout etc as a db value ?

I don't think we need to worry about the timeout exactly, but I'd expect DynamoDB to have records _something_ like:
| userId | evidence | ttl |
| --- | --- | --- |
| urn:uuid:some-user-id | `{ "type": "RiskAssessment" }` | 1700667070 |
| urn:uuid:some-other-id | `{ "type": "RiskAssessment", "ci": ["XXX"], "txt": "some-txn-id" }` | 1700667070 |

I'll raise a separate ticket for this

di-ipv-ticf-stub/lambdas/.eslintrc.js
Joe-Edwards-GDS, line 1: I think this still needs updating to use the typescript plugin - have a look at the example in the delete-user-data lambda.

di-ipv-ticf-stub/lambdas/package.json
Joe-Edwards-GDS: Can you elaborate on this? I can't see any usages of it in the code

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
```diff
@@ -0,0 +1,20 @@
+export const apiResponses = {
+  _200: (body: { [key: string]: any }) => ({
```
Joe-Edwards-GDS: I think you could just use `body: object` here?

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+import ticfService from "../services/ticfService";
+
+export const handler: Handler = async (event: any) => {
+  let ticfRequest: TicfRequest;
+  try {
+    ticfRequest = JSON.parse(event.body);
+  } catch (error) {
+    return apiResponses._400();
+  }
+  if (ticfRequest == null || ticfRequest.vot == null)
```
Joe-Edwards-GDS: You should always use `===` in typescript, unless you _really_ want type coercion.

In this case I don't think these would ever be `null`, but it's perfectly idiomatic to use falsiness - assuming that 'empty' is invalid:
```
if (!ticfRequest || !ticfRequest.vot) {
    return apiResponses._400({ errorMessage: "Invalid request" });
  }
```
Should we also check the other properties of the request are present too?

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
Joe-Edwards-GDS: It would be nice to specify the return types here - I think it's `ApiGatewayProxyResultV2`?

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
```diff
@@ -0,0 +1,20 @@
...
+    statusCode: 200,
+    body: JSON.stringify(body, undefined, 2),
```
Joe-Edwards-GDS, line 7: Do you need to include a content type here - i.e. `headers: { 'content-type': 'application/json' },`?

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+import { v4 as uuid } from "uuid";
+
+import { buildSignedJwt } from "stub-oauth-client";
+import type { SignedJwtParams } from "stub-oauth-client";
+
+import { getSsmParameter } from "../common/ssmParameter";
+import TicfRequest from "../domain/ticfRequest";
+import TicfResponse from "../domain/ticfResponse";
+
+export class TicfService {
```
Joe-Edwards-GDS: I don't think this needs to be a class - it doesn't have any state. In general I prefer quite functional Typescript anyway*, and I think this could just be a direct export:

```
export async function processGetVcRequest(ticfRequest: TicfRequest): Promise<TicfResponse> {
  ...
}
```
*JavaScript classes are _weird_ and full of gotchas

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+  async processGetVCRequest(ticfRequest: TicfRequest): Promise<TicfResponse> {
+    // preparing response
+    let ticfSigningKey: string;
+    try {
+      ticfSigningKey = await getSsmParameter(
+        process.env.TICF_PARAM_BASE_PATH + "signingKey"
+      );
+    } catch (error) {
+      throw new Error(`Error while retrieving TicF CRI VC signing key.`);
+    }
```
Joe-Edwards-GDS: This is swallowing the underlying error - can we just let that bubble up? An error message describing the parameter being fetched feels more useful anyway.

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+    let timeoutVC: string | null | undefined = await getSsmParameter(
```
Joe-Edwards-GDS: You shouldn't need the explicit type annotation on this, it should be inferred

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+      process.env.TICF_PARAM_BASE_PATH + "timeoutVC"
+    );
+    timeoutVC ??= "false";
```
Joe-Edwards-GDS: I don't think this can ever happen, now that `getSsmParameter` throws for an undefined parameter

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+    let includeCIToVC: string | null | undefined = await getSsmParameter(
+      process.env.TICF_PARAM_BASE_PATH + "includeCIToVC"
+    );
+    includeCIToVC ??= "false";
+
+    let buildJwtParams: SignedJwtParams;
+    buildJwtParams = {
```
Joe-Edwards-GDS: Can just be
```
const buildJwtParams = {
  ...
};
```

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+        JSON.parse(timeoutVC.toLowerCase()),
+        JSON.parse(includeCIToVC.toLowerCase()),
+        ticfRequest.sub
+      ),
+      privateSigningKey: ticfSigningKey!,
+    };
+
+    // preparing response
+    const returnJwt = await buildSignedJwt(buildJwtParams);
+    let responseBody: TicfResponse;
```
Joe-Edwards-GDS: Can just be
```
const returnJwt = await buildSignedJwt(buildJwtParams);

return {
  sub: ticfRequest.sub,
  ...
};
```

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+function getCustomClaims(
+  timeOutVC: boolean,
+  includeCIToVC: boolean,
+  userId: string
+): JWTPayload {
+  if (timeOutVC) {
+    return {
+      sub: userId,
+      iss: process.env.ISSUER,
+      aud: "https://development-di-ipv-core-front.london.cloudapps.digital",
```
Joe-Edwards-GDS: Do you think this needs to be configurable as well? How does it work for the other stubs?

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -0,0 +1,120 @@
...
+              txn: uuid(),
+            },
+          ],
+          type: ["VerifiableCredential", "RiskAssessmentCredential"],
+        },
+        jti: "urn:uuid:" + uuid(),
+      };
+    }
+  }
+}
```
Joe-Edwards-GDS, line 108: There's still repetition here, that I think we can avoid?
```
function getCustomClaims(
  timeOutVC: boolean,
  includeCIToVC: boolean,
  userId: string
): JWTPayload {
  return {
    sub: userId,
    iss: process.env.ISSUER,
    aud: "https://development-di-ipv-core-front.london.cloudapps.digital",
    nbf: Date.now(),
    vc: {
      evidence: [getEvidenceItem(timeOutVC, includeCIToVC)],
      type: ["VerifiableCredential", "RiskAssessmentCredential"],
    },
    jti: "urn:uuid:" + uuid(),
  };
}

function getEvidenceItem(
  timeOutVC: boolean,
  includeCIToVC: boolean
): TicfEvidenceItem {
  if (timeOutVC) {
    return { type: "RiskAssessment" };
  } else {
    if (includeCIToVC) {
      return {
        type: "RiskAssessment",
        ci: ["V03"],
        txn: uuid(),
      };
    } else {
      return {
        type: "RiskAssessment",
        txn: uuid(),
      };
    }
  }
}
```
This should also make it easy to update when the evidence block comes from the management API

di-ipv-ticf-stub/lambdas/src/domain/ticfResponse.ts
Joe-Edwards-GDS, line 1: (If these types start proliferating, then I think it'd also be fine to put them in a single file)

di-ipv-ticf-stub/lambdas/.eslintrc.js
```diff
@@ -1,22 +1,23 @@
...
+  globals: {
+    expect: true,
+  },
+  root: true,
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  rules: {
+    "no-console": 0,
+    "@typescript-eslint/no-explicit-any": ["off"],
```
Joe-Edwards-GDS: Can we try to keep this if possible - where do you need to use `any`?

di-ipv-ticf-stub/lambdas/package.json
rpayal: in service using - process.env.TICF_PARAM_BASE_PATH etc.   when removing ts-node it was complaining.

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
rpayal: check added for other properties also.

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
rpayal: already taken care

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
rpayal: done

di-ipv-ticf-stub/lambdas/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "babel-jest": "^29.7.0",
+    "eslint": "^8.54.0",
+    "jest": "^29.7.0",
+    "typescript": "^5.3.2"
+  },
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w",
+    "test": "jest --config=jest.config.test.js",
+    "integ-test": "jest --config=jest.confg.integ.test.js",
```
Joe-Edwards-GDS: Are you planning to write integration tests for this?

di-ipv-ticf-stub/lambdas/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "aws-sdk": "^2.1497.0",
+    "jose": "4.11.1",
+    "stub-oauth-client": "github:govuk-one-login/stub-oauth-client#main",
+    "ts-node": "^10.9.1"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.23.3",
+    "@babel/preset-env": "^7.23.3",
+    "@babel/preset-typescript": "^7.23.3",
+    "@types/axios": "^0.14.0",
```
Joe-Edwards-GDS: It looks like axios provides its own types now, so you might not need this

di-ipv-ticf-stub/lambdas/test/unit/ticfHandler.test.ts
Joe-Edwards-GDS, line 1: If you're having type errors in your test, you might need to tell tsconfig to include them, which either means:
- Removing the exclusion on `node_modules` (so it includes the Jest types by default)
- Including them explicitly, with `"types": "jest"`

di-ipv-ticf-stub/lambdas/package.json
rpayal: will remove - earlier added was trying to create integration test agnst local run.

di-ipv-ticf-stub/openAPI/ticf-external.yaml
rpayal: https://docs.aws.amazon.com/apigateway/latest/developerguide/integration-passthrough-behaviors.html
- The request payload is transformed using the template.
Changes application/x-www-form-urlencoded to application/json

Joe-Edwards-GDS: Do we need this at all though? I think we can just pass through directly to the lambda

di-ipv-ticf-stub/lambdas/src/common/apiResponses.ts
rpayal: Done

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
rpayal: done

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
rpayal: done

Joe-Edwards-GDS (dismissed):
Happy to get this unblocked now :)

I will probably raise another PR to tidy up a few extra bits and pieces (e.g. sorting out this `di-stub-oauth-client` issue), but no need to delay any further.

di-ipv-ticf-stub/lambdas/package.json
rpayal: used for integ test - removed now.

di-ipv-ticf-stub/lambdas/.eslintrc.js
rpayal: changed to 'error'

rpayal:
Tested in dev.

Joe-Edwards-GDS (dismissed):

di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
rpayal: Updated to get componentId from ssm for VC's aud value.

Joe-Edwards-GDS (approved):

PR #576, opened 2023-11-24
Pyic 3894 tidy up

Joe-Edwards-GDS:
Tidy up various bits of code in the TICF CRI stub

### DIFF
--- a/di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
+++ b/di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
@@ -3,6 +3,7 @@ import { getSsmParameter } from "../common/ssmParameter";
 import TicfRequest from "../domain/ticfRequest";
 import TicfResponse from "../domain/ticfResponse";
 import TicfEvidenceItem from "../domain/ticfEvidenceItem";
+import TicfVc from "../domain/ticfVc";
 import { signJwt } from "./signingService";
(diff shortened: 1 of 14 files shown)


di-ipv-ticf-stub/lambdas/src/services/ticfService.ts
```diff
@@ -1,98 +1,53 @@
...
-): JWTPayload {
-  return {
-    sub: userId,
-    iss: process.env.ISSUER,
-    aud: componentId,
-    nbf: Date.now(),
+  const payload: TicfVc = {
+    iss: ticfComponentId,
+    sub: ticfRequest.sub,
+    aud: ticfComponentId,
```
Joe-Edwards-GDS, line 22: I'm not sure the audience should actually use the same value as the issuer, but I'm checking with David to see what the right thing to do here is. I think the audience should actually be the IPV Core component id.

rpayal: Issuer will be ticf in this case (they r the one issuing this VC)....audience will be ipv core as they r the one receiving it.
e.g aud - "https://development-di-ipv-core-front.london.cloudapps.digital"
  iss - "https://ticf-cri.stubs.account.gov.uk"

rpayal (approved):
LGTM

PR #580, opened 2023-11-30
BAU: Rename main deploy template export name

Wynndow:
### What changed

Rename main deploy template export name

### Why did it change

This is the template used by the pipeline. It clashes with another export name and needs updating.


rpayal (approved):
oh....I did in dev one.,...missd this one.

PR #585, opened 2023-12-11
BAU: update to govuk-one-login

jkunle:
### What changed

Organisation name

### Why did it change

Organisation name has changed to  `govuk-one-login`

### DIFF
--- a/.github/workflows/core-cri-preprod-stub.yml
+++ b/.github/workflows/core-cri-preprod-stub.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Checkout config repo
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


merlinc (dismissed):

surajkumar (changes requested):
Might as well update the SAM deploy to use `govuk-one-login/devplatform-upload-action@v3.3`

Can use this for ref: https://github.com/govuk-one-login/ipv-cri-check-hmrc-api/blob/c04b297a08583ac21a2daa8d35636fe8c295b850/.github/workflows/post-merge-deploy-to-build.yml#L43C11-L43C11

surajkumar (approved):

PR #586, opened 2023-12-12
PYIC 4188: added a new property 'issuers' in the contraIndicator and GetContraIndicatorCredential Lambda's response in the cimit-stub

TalatAydinGDS:
### What changed

added a new property 'issuers' in the contraIndicator and GetContraIndicatorCredential Lambda's response in the cimit-stub

### Why did it change

The GetContraIndicatorCredential Stub Lambda should return issuers from response like Cimit service

### Issue tracking

- [PYIC-4188](https://govukverify.atlassian.net/browse/PYIC-4188)

### Other considerations

[PYIC-4188]: https://govukverify.atlassian.net/browse/PYIC-4188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (commented):
How complex do you think it would be to store the actual issuer received from the `PutContraIndicators` endpoint, rather than just hardcoding to the two examples?

We should have the issuer in the VC submitted to CIMIT, so we might be able to just add it to the `CimitStubItem`...

(This will be a lot closer to how CIMIT actually works, and should make it more representative for testing as well)

TalatAydinGDS:
We have implemented an alternative solution of a configuration change to be more generic, and we are closing it for this reason.

PR #596, opened 2024-01-02
BAU: Fetch client config from SSM

Wynndow:
**Not to be merged before https://github.com/govuk-one-login/ipv-stubs-common-infra/pull/150**

### What changed

Fetch client config from SSM

### Why did it change

This creates a new config service for fetching client config from the parameter store, rather than loading it in from a file.

This moves us towards how we manage config in the other components, and should make the set up for the CRI stub easier to understand.

The new ConfigService is currently only used for client config, but could be extended for all config in future work.

It's been implemented as a static class. This means it can just be called where it's needed and avoids having to inject it everywhere. This does make it harder to test, and a stub client has been created to allow some testing to happen.


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/ConfigService.java
```diff
@@ -0,0 +1,87 @@
...
+
+        return CLIENT_CONFIGS;
+    }
+
+    private static void refreshClientConfigsIfRequired() {
+        if (lastRefresh == null
+                || lastRefresh.isBefore(Instant.now().minusSeconds(getCacheDurationSeconds()))) {
+            LOGGER.info("Refreshing client configs");
+            getAllClientConfigs(null);
+            lastRefresh = Instant.now();
```
Joe-Edwards-GDS, line 47: This is not thread-safe - do you think it needs to be?

In the most common cases it will probably just mean the config gets refreshed twice, but in the worst case it could corrupt the hash map.

You could avoid the latter case by making `getAllClientConfigs` build a fresh map, rather than mutating the original.

di-ipv-credential-issuer-stub/src/test/java/uk/gov/di/ipv/stub/cred/fixtures/TestFixtures.java
```diff
@@ -29,9 +24,58 @@ private TestFixtures() {}
...
+
+    public static final String CLIENT_CONFIG =
```
Joe-Edwards-GDS: Could this just be a `ClientConfig` object, to avoid serialising/deserialising? I guess it'll be slightly clunkier in Java syntax :/

If not, I wonder if it'd be cleaner to load this as a JSON resource file, rather than embedding in Java.

Joe-Edwards-GDS (dismissed):
I'm not really a fan of the static singleton, but I can appreciate that it's much easier to implement this way(!)

Happy to leave it like this for now as it's no worse than the previous code, but left a couple of comments you might want to consider as well.

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/ConfigService.java
Wynndow: Hadn't thought about threading - even though I don't think we need it right now it's probably a good idea to deal with it in case we do in future.

I've not done loads of multithreaded stuff, but I'm tempted to make the `refreshClientConfigsIfRequried` method `synchronized`. I think that should allow multiple threads to use the same `lastRefresh` timestamp - which would be advantageous, as well as protect the map? I'll update `getAllClientConfigs` to generate a new map rather than mutate too though.

Interested to understand how the hash map might get corrupted in the current implementation though?

di-ipv-credential-issuer-stub/src/test/java/uk/gov/di/ipv/stub/cred/fixtures/TestFixtures.java
Wynndow: I think making it an object will be ok - will try it. I changed it from being b64 encoded to make it readable, but should have gone a step further!

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/ConfigService.java
Joe-Edwards-GDS: Yeah, one neat thing about lambdas is that we usually don't have to worry about it (unless we deliberately write multithreaded code), but if this is running in a container and shared across requests, we probably do.

Making the method synchronised means that only one thread can enter the method at a time (and threads will just block until they can). In complex scenarios that can risk deadlocks etc. but that shouldn't be a worry here.

The issue previously is that if multiple threads attempt to read/update the map at the same time, you can get race conditions in the internal workings of the hashmap. I'm a bit rusty on all the details, but they usually have 'buckets' of hashes, and as you add/remove items, these can change in size. If a thread reads/updates during one of these operations, it could potentially lead to items getting lost, getting stale/unexpected values, etc.

Joe-Edwards-GDS (dismissed):

di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/service/ConfigService.java
Wynndow: Cool, makes sense. Thanks Joe

Joe-Edwards-GDS (approved):

PR #604, opened 2024-01-11
PYIC-4262: Enhance Orchestration Stub to provide Inherited Id's Claim and VTR

TalatAydinGDS:
### What changed

Update the Orchestration Stub to allow us to simulate a user migrating to One Login

### Why did it change

HMRC will send a new claim to us that will contain information about the user that is migrating their Government Gateway account over to OneLogin. 

This information will include their current VOT (according to HMRC), Their name, dob, and social security details and An HMRC signed VC that may or may not contain evidence of them having completed the live/likeness application to achieve a current PCL250 status. 

### Issue tracking

- [PYIC-4262](https://govukverify.atlassian.net/browse/PYIC-4262)

### Other considerations

[PYIC-4262]: https://govukverify.atlassian.net/browse/PYIC-4262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


di-ipv-orchestrator-stub/deploy-integration/template.yaml
```diff
@@ -64,6 +64,22 @@ Parameters:
     Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
     Type: AWS::SSM::Parameter::Value<String>
     Default: "/stubs/integration/orch/env/ORCHESTRATOR_BASIC_AUTH_PASSWORD" #pragma: allowlist secret
+  InheritedIdentityJwtSigningKey:
+    Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: "/stubs/build/orch/env/INHERITED_IDENTITY_JWT_SIGNING_KEY" #pragma: allowlist secret
```
Wynndow: I think the default for these params will need to start with `/stubs/integration...`

di-ipv-orchestrator-stub/deploy-staging/template.yaml
```diff
@@ -64,6 +64,22 @@ Parameters:
     Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
     Type: AWS::SSM::Parameter::Value<String>
     Default: "/stubs/staging/orch/env/ORCHESTRATOR_BASIC_AUTH_PASSWORD" #pragma: allowlist secret
+  InheritedIdentityJwtSigningKey:
+    Description: "Environment Variables for ECS - Retrieved from SSM Parameter Store"
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: "/stubs/build/orch/env/INHERITED_IDENTITY_JWT_SIGNING_KEY" #pragma: allowlist secret
```
Wynndow: Same here, will need to start `/stubs/staging/...`

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
```diff
@@ -33,6 +33,16 @@ public class OrchestratorConfig {
     public static final String BASIC_AUTH_PASSWORD =
             getConfigValue("ORCHESTRATOR_BASIC_AUTH_PASSWORD", null);
 
+    public static final String INHERITED_IDENTITY_JWT_SIGNING_KEY =
+            getConfigValue("INHERITED_IDENTITY_JWT_SIGNING_KEY", "missing-key");
```
Wynndow: might be better to use `null` rather than `"missing-key"`

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/HomeHandler.java
```diff
@@ -21,7 +24,34 @@ public class HomeHandler {
...
-
+                moustacheDataModel.put("credentialSubjects", getCredentialSubjectData());
+                moustacheDataModel.put("evidences", getEvidencePayloads());
                 return ViewHelper.render(moustacheDataModel, "home.mustache");
             };
+
+    private static Object getCredentialSubjectData()
+            throws UnsupportedEncodingException,
+                    com.nimbusds.jose.shaded.json.parser.ParseException {
+        com.nimbusds.jose.shaded.json.JSONObject js =
```
Wynndow: Could these classes be imported directrly, rather than using their fully qualified names here?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/HomeHandler.java
```diff
@@ -21,7 +24,34 @@ public class HomeHandler {
...
+                                        HomeHandler.class.getResourceAsStream(
+                                                "/data/inheritedJWTCredentialSubjects.json"));
+
+        return js.get("data");
+    }
+
+    private static Object getEvidencePayloads()
+            throws UnsupportedEncodingException,
+                    com.nimbusds.jose.shaded.json.parser.ParseException {
+        com.nimbusds.jose.shaded.json.JSONObject js =
```
Wynndow: Same here

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -89,16 +92,24 @@ public class IpvHandler {
                                         reproveIdentityString)
                                 : JwtBuilder.ReproveIdentityClaimValue.NOT_PRESENT;
 
+                String credentialSubject = request.queryMap().value(JSON_PAYLOAD_PARAM);
```
Wynndow, line 100: Could this param be something more descriptive - maybe `credentialSubjectJson`?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -89,16 +92,24 @@ public class IpvHandler {
...
+                        Objects.equals(request.queryMap().value(DURING_MIGRATION), "checked");
+
                 String userId = getUserIdValue(userIdTextValue);
 
                 JWTClaimsSet claims =
                         JwtBuilder.buildAuthorizationRequestClaims(
                                 userId,
                                 signInJourneyIdText,
-                                vtr,
+                                vtr[0],
```
Wynndow: I think we can have multiple vtr values, should we pass the whole array in?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/HMRCUserInfoResponse.java
```diff
@@ -0,0 +1,39 @@
...
+            throws NoSuchAlgorithmException, InvalidKeySpecException, JOSEException {
+        UserInfo userInfo = new UserInfo();
+        userInfo.setInheritedIdentityJWT(
+                duringMigration
+                        ? new InheritedIdentityJWT(
+                                InheritedIdentityJWTBuilder.generate(
+                                                userId, vtr, credentialSubject, evidence)
+                                        .serialize())
+                        : null);
+        ObjectMapper objectMapper = new ObjectMapper();
```
Wynndow, line 36: This objectMapper could be moved to a static field to avoid creating a new one every time this method is called.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+        vc.put(VC_CREDENTIAL_SUBJECT, convertJsonToMap(credentialSubject));
+        vc.put(VC_EVIDENCE, List.of(convertJsonToMap(evidence)));
+        Instant now = Instant.now();
+        JWTClaimsSet claimsSet =
+                new JWTClaimsSet.Builder()
+                        .subject(userId)
+                        .issuer(INHERITED_IDENTITY_JWT_ISSUER)
+                        .notBeforeTime(Date.from(now))
+                        .expirationTime(generateExpirationTime(now))
+                        .claim(VOT, vtr)
```
Wynndow: It might be good at some point to make the VOT in the VC not coupled to the requested vtr. That way we can test scenarios where they're requestion a VOT that doesn't match what's in the inherited identity.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+                        .claim(VTM, INHERITED_IDENTITY_JWT_VTM)
+                        .claim(VC, vc)
+                        .build();
+        return createSignedJwt(claimsSet);
+    }
+
+    private static Map<String, Object> convertJsonToMap(String json) {
+        try {
+            return new ObjectMapper().readValue(json, new TypeReference<>() {});
```
Wynndow: Might be worth making a static object mapper that can be reused rather than generating a new one very time.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JwtBuilder.java
```diff
@@ -85,8 +92,8 @@ public static JWTClaimsSet buildAuthorizationRequestClaims(
                         .claim("govuk_signin_journey_id", signInJourneyId)
                         .claim("persistent_session_id", UUID.randomUUID().toString())
                         .claim("email_address", userEmailAddress)
-                        .claim("vtr", List.of(vtr));
-
+                        .claim("vtr", List.of(vtr))
+                        .claim("userInfo", userInfo);
```
Wynndow: I think that the `userinfo` claim is nested inside a top level `claims` claim.

di-ipv-orchestrator-stub/src/main/resources/data/inheritedJWTCredentialSubjects.json
```diff
@@ -0,0 +1,37 @@
+{
+  "data": [
+    {
+      "label": "Alice Jane Dow",
```
Wynndow: Might be worth adding Kenneth too.

I don't think it should actually matter - we won't be correlating the names, but it might prove useful and less confusing for testing purposes if a journey can be run with the same user.

di-ipv-orchestrator-stub/src/main/resources/data/inheritedJWTEvidences.json
```diff
@@ -0,0 +1,42 @@
+{
+  "data": [
+    {
+      "scenario": "Cohort1",
+      "label": "Users with VoT PCL200 or PCL250, with no evidence provided",
+      "payload": {
```
Wynndow, line 6: Should the value of this payload be `null`? The example shows that it should be an empty array in the final VC, and the value of this payload gets set in an array when constructing the VC.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
TalatAydinGDS: Valid values("PCL200", "PCL250", "P2") are checked within the list and retrieved, and a warning message is added to the log for the others.

TalatAydinGDS: Chris, I have a question: If there are more than one occurrence of values 'PCL200', 'PCL250', 'P2' within the VTR array, should we write them as separate VOT array values, rather than within a single VOT entry? Although it may not be feasible from a business perspective, I am considering this possibility based on the data.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
```diff
@@ -47,6 +47,16 @@ public class OrchestratorConfig {
     public static final String BASIC_AUTH_PASSWORD =
             getConfigValue("ORCHESTRATOR_BASIC_AUTH_PASSWORD", null);
 
+    public static final String INHERITED_IDENTITY_JWT_SIGNING_KEY =
+            getConfigValue("INHERITED_IDENTITY_JWT_SIGNING_KEY", null);
+    public static final String INHERITED_IDENTITY_JWT_ISSUER =
+            getConfigValue(
+                    "INHERITED_IDENTITY_JWT_ISSUER", "<https://oidc.hmrc.gov.uk/migration/v1>");
+    public static final String INHERITED_IDENTITY_JWT_VTM =
+            getConfigValue("INHERITED_IDENTITY_JWT_VTM", "<https://hmrc.gov.uk/trustmark>");
```
Joe-Edwards-GDS: The angle brackets were meant to indicate that these are unconfirmed, but I think we can:
- Use a stub issuer value (similar to other CRI stubs), e.g. `https://orch.stubs.account.gov.uk/migration/v1` - this will need to match the stub config in core-back (see: https://github.com/govuk-one-login/ipv-core-common-infra/pull/819/files)
- Use the expected vtm value, i.e. `https://hmrc.gov.uk/trustmark` (with no angle brackets)

cc @MikeCollingwood

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/HomeHandler.java
```diff
@@ -21,7 +25,20 @@ public class HomeHandler {
...
-
+                moustacheDataModel.put(
+                        "credentialSubjects", getData("/data/inheritedJWTCredentialSubjects.json"));
+                moustacheDataModel.put("evidences", getData("/data/inheritedJWTEvidences.json"));
                 return ViewHelper.render(moustacheDataModel, "home.mustache");
             };
+
+    private static Object getData(String jsonPath)
+            throws UnsupportedEncodingException,
+                    com.nimbusds.jose.shaded.json.parser.ParseException {
```
Joe-Edwards-GDS: Do we need the FQCN here?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/HomeHandler.java
```diff
@@ -21,7 +25,20 @@ public class HomeHandler {
...
+        JSONObject js =
+                (JSONObject)
+                        new JSONParser(MODE_JSON_SIMPLE)
+                                .parse(HomeHandler.class.getResourceAsStream(jsonPath));
```
Joe-Edwards-GDS: If you've added the Jackson library, should we use that here for consistency?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JwtBuilder.java
```diff
@@ -82,6 +89,7 @@ public static JWTClaimsSet buildAuthorizationRequestClaims(
                         .issuer(ORCHESTRATOR_CLIENT_ID)
                         .notBeforeTime(Date.from(now))
                         .expirationTime(generateExpirationTime(now))
+                        .claim("userInfo", userInfo)
```
Joe-Edwards-GDS: Sorry Talat, this wasn't clear in the original ticket but I think this will actually be wrapped in a `claims` object, i.e.:
```
{
  "sub": "...",
  "iss": "...",
  ...
  "claims": {
    "userinfo": {
      "https://vocab.account.gov.uk/v1/inheritedIdentityJWT": {
        "values": ["<base64 encoded VC"]
      }
    }
  }
}
```
Hope that's not too hard to add (it's based on https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter)

di-ipv-orchestrator-stub/src/main/resources/data/inheritedJWTCredentialSubjects.json
```diff
@@ -0,0 +1,66 @@
...
```
Joe-Edwards-GDS: Typo: 'Doe'

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
```diff
@@ -89,6 +91,25 @@
                 <input class="govuk-input" data-module="govuk-input" name="emailAddress" id="emailAddress" type="text" value="dev-platform-testing@digital.cabinet-office.gov.uk">
             </div>
 
+            <div class="govuk-checkboxes" data-module="govuk-checkboxes">
+              <div class="govuk-checkboxes__item">
+                <input class="govuk-checkboxes__input" type="checkbox" name="duringMigration" id="duringMigration" value="checked">
+                <label class="govuk-label govuk-checkboxes__label" for="duringMigration">
+                During Migration
```
Joe-Edwards-GDS, line 102: Could we rename this to be a bit clearer - perhaps 'Include inherited identity' or similar?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+                !evidence.equals("{}")
+                        ? List.of(convertJsonToMap(evidence))
+                        : Collections.emptyList());
+        Instant now = Instant.now();
+        JWTClaimsSet claimsSet =
+                new JWTClaimsSet.Builder()
+                        .subject(userId)
+                        .issuer(INHERITED_IDENTITY_JWT_ISSUER)
+                        .notBeforeTime(Date.from(now))
+                        .expirationTime(generateExpirationTime(now))
```
Joe-Edwards-GDS, line 72: I don't think our credentials have an expiry time any more, so we can probably just remove this

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+                        .claim(VOT, getValidVtrValues(vtr))
```
Joe-Edwards-GDS: This isn't right - the VOT on the credential should be a single value, and is independent from the VTR on the JAR. You'll probably need another field on the stub to enter the VOT for the inherited ID.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+                        .claim(VTM, INHERITED_IDENTITY_JWT_VTM)
+                        .claim(VC, vc)
+                        .build();
```
Joe-Edwards-GDS, line 81: I think we should also add a JTI to this - it's not in the example, but it is in our spec for VCs (https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3834876636/RFC+0024+Identity+RFC+0024+Verifiable+Credentials+for+Identity+Checks#Verifiable-credential-structure)

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
Joe-Edwards-GDS: It's legitimate to have a VTR = ["PCL250", "P2"], and a VOT = "PCL200"

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+                .toList();
+
+        return Arrays.stream(vtr).filter(allowedValues::contains).collect(Collectors.toList());
+    }
+
+    private static Map<String, Object> convertJsonToMap(String json) {
+        try {
+            return objectMapper.readValue(json, new TypeReference<>() {});
+        } catch (JsonProcessingException e) {
+            return Collections.emptyMap();
```
Joe-Edwards-GDS: Are you sure we want to just return an empty map here - is that expected behaviour?

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
```diff
@@ -89,6 +91,25 @@
...
+                </label>
+              </div>
+            </div><br/>
+            <div class="govuk-form-group">
+                <label class="govuk-label" for="vtrText">Enter vtr manually</label>
+                <input class="govuk-input" data-module="govuk-input" name="vtrText" id="vtrText" type="text" value="P2">
+            </div>
```
Joe-Edwards-GDS: The VTR may be useful for testing HMRC journeys _after_ migration as well (i.e. without an inherited identity), so we may want to move this back up.

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
```diff
@@ -89,6 +91,25 @@
...
+
+            <div class="govuk-form-group" id="test_data_block"></div>
+            <div class="govuk-form-group"><textarea class="govuk-textarea" id="jsonPayload" name="jsonPayload" rows="5" aria-describedby="more-detail-hint"></textarea></div>
+
+            <div class="govuk-form-group" id="custom_evidence_block"></div>
+            <div class="govuk-form-group"> <textarea class="govuk-textarea" id="evidenceJsonPayload" name="evidenceJsonPayload" rows="5" aria-describedby="more-detail-hint"></textarea></div>
```
Joe-Edwards-GDS: I wonder if it'd be convenient to add a bit of JS that hides/shows these two fields depending on whether you tick the inherited ID checkbox... What do you think?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
TalatAydinGDS: created a JWT based on the example provided in the task. Could you please add an example JWT with correct values for all fields to the task? The reason I requested a JWT containing all fields was because of our previous discussion on the Slack channel.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/HomeHandler.java
TalatAydinGDS: I took this code directly from the di-ipv-credential-issuer-stub. It will also need corrections in the future. I'll be updating here.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
TalatAydinGDS: Joe, I couldn't see this detail in the task description. I wish I could have seen this detail within the task before raising the PR. If validation is required for the change, and if an exception is expected at the end of it, please specify.

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
TalatAydinGDS: The VOT has been added as a separate entry on the front end, thus reverting the VTR to its previous location.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
Joe-Edwards-GDS: I don't think there's a complete example of a 'valid' VC yet, because a few of the details are TBD, and the examples in the RFC are for prod (not the stub). I think it should be roughly:
```
{
  "sub": "urn:fdc:gov.uk:2022:A_w9_Or3gyI-ygrzH7cNZe9qTYFkUOzHF-KoPWmb9QY",
  "iss": "https://orch.stubs.account.gov.uk/migration/v1",
  "jti": "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
  "nbf": 1541493724,
  "vot": "CL200",
  "vtm": "https://hmrc.gov.uk/trustmark",
  "vc": {
    "type": [
      "VerifiableCredential",
      "IdentityCheckCredential"
    ],
    "credentialSubject": {
      "name": [
        {
          "nameParts": [
            {
              "value": "Alice",
              "type": "GivenName"
            },
            {
              "value": "Jane",
              "type": "GivenName"
            },
            {
              "value": "Laura",
              "type": "GivenName"
            },
            {
              "value": "Doe",
              "type": "FamilyName"
            }
          ]
        }
      ],
      "birthDate": [
        {
          "value": "1970-01-01"
        }
      ],
      "socialSecurityRecord": [
        {
          "personalNumber": "AB123456C"
        }
      ]
    }
  },
  "evidence": [
    {
      "type": "IdentityCheck",
      "txn": "d22f8cb0-82f2-4526-9d88-07ccd358ab13",
      "strengthScore": 2,
      "validityScore": 2,
      "verificationScore": 1
      "checkDetails": [
        {
          "checkMethod": "data"
        }
      ]
    }
  ]
}
```
This is based on the draft RFC (https://github.com/govuk-one-login/architecture/pull/399/files), adding the JTI and updating the issuer to reflect that it's from a stub.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
Joe-Edwards-GDS: Sorry it wasn't clear, the two fields have different meanings:
- VTR = Vector of Trust request, specified on the JAR to indicate the level of confidence that the RP requires us to prove an identity to
- VOT = Vector of Trust, specified on the VC to indicate the level of confidence that the VC can prove

I don't think we need exception handling - all the handling logic happens in IPV Core, the stub should just pass whatever is specified.

di-ipv-orchestrator-stub/src/main/resources/templates/home.mustache
TalatAydinGDS: After the 'inheritedidentity jwt during migration' checkbox is checked, it is linked to the creation process. I don't think there is a need to add a new checkbox. If that one is selected, the relevant fields are displayed.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
TalatAydinGDS: I didn't know this, unfortunately. Could you please provide an example of a JWT that includes this field and all other fields in this area?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/models/InheritedIdentityJWT.java
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class InheritedIdentityJWT {
+    private String[] value;
```
Joe-Edwards-GDS: Sorry Talat, it looks like this was updated earlier today to be `values` (not `value`). I know this must be frustrating!

https://github.com/govuk-one-login/architecture/pull/554/files

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
TalatAydinGDS: Thank you.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
TalatAydinGDS: Ok. Thank you

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/InheritedIdentityJWTBuilder.java
Joe-Edwards-GDS: Yep, I think the RFC has now been updated to include JTI: https://github.com/govuk-one-login/architecture/pull/554/files (I've also put a full example in another comment)

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
rpayal: @Joe-Edwards-GDS 
Since as part of (https://govukverify.atlassian.net/browse/PYIC-4339) while preparing user identity for operational profile VOT we are only including operational vc, ticf and cimit vc. 
So I think this VC should also include "address" array block under "credentialSubject" ?

Joe-Edwards-GDS: @rpayal good question

We don't expect to receive an address in the inherited identity VC - and as you've spotted, this does mean we won't return an address in the user identity when returning an operational profile.

I haven't explicitly considered the impact of this - if this is going to break something we might need to soften our handling of a missing address.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/utils/JwtBuilder.java
MikeCollingwood: Is it "values" now?
```
...
"https://vocab.account.gov.uk/v1/inheritedIdentityJWT": {
    "values": ["<base64 encoded VC"]
}
...
```

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
rpayal: @Joe-Edwards-GDS Currently ProvenUserIdentity take user to journey error path if address is not found.

Joe-Edwards-GDS: Ok thanks - we probably need a new ticket for this (unless you want to add it to PYIC-4339)

TalatAydinGDS: changed iss value with 'https://orch.stubs.account.gov.uk/migration/v1' according above jwt example. But, can be updated by modifying the parameter INHERITED_IDENTITY_JWT_ISSUER in the configuration.

Joe-Edwards-GDS (approved):
When it starts being used a bit more, we can determine whether any of the fields need updating.

PR #606, opened 2024-01-16
PYIC-4461 Increase orchestrator stub capacity

Joe-Edwards-GDS:
In the latest load test the orchestrator maxed out at 100% and became a bottleneck for the test.

The orchestrator stub does not appear to use any state, so we should be able to safely scale it horizontally. Ideally this would happen with an autoscaling configuration, but we want to try it first.

### DIFF
--- a/di-ipv-orchestrator-stub/deploy/template.yaml
+++ b/di-ipv-orchestrator-stub/deploy/template.yaml
@@ -153,9 +153,9 @@ Mappings:
       platform: stubs
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
     "388905755587": #Stubs Prod
-      fargateCPUsize: "4096"
-      fargateRAMsize: "8192"
-      desiredTaskCount: 1
+      fargateCPUsize: "2048"
+      fargateRAMsize: "4096"
+      desiredTaskCount: 4
       environment: production
       platform: stubs
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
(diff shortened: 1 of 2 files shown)


di-ipv-orchestrator-stub/deploy/template.yaml
```diff
@@ -153,9 +153,9 @@ Mappings:
       platform: stubs
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
     "388905755587": #Stubs Prod
-      fargateCPUsize: "4096"
-      fargateRAMsize: "8192"
-      desiredTaskCount: 1
+      fargateCPUsize: "2048"
+      fargateRAMsize: "4096"
+      desiredTaskCount: 4
```
Joe-Edwards-GDS, line 158: Back of the envelope says this will change run cost from ~£100/month -> ~£200/month

We might want autoscaling or a way to manually scale for the duration of a load test, but I thought it better to try this first.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -67,12 +67,10 @@ public class IpvHandler {
             "https://vocab.account.gov.uk/v1/credentialJWT";
 
     private final Logger logger = LoggerFactory.getLogger(IpvHandler.class);
-    private final Map<String, Object> stateSession = new HashMap<>();
 
     public Route doAuthorize =
             (Request request, Response response) -> {
                 var state = new State();
-                stateSession.put(state.getValue(), null);
```
Wynndow, line 75: Sort of feel like we should fix this so we are checking the correct state is being returned. If we broke it somehow, we wouldn't know until we hit prod.

Looks like [we originally were checking](https://github.com/govuk-one-login/ipv-stubs/pull/7/commits/778028783422f1f84d004ef9de35bdfd2d55b514#diff-4b5658ef0c7b2bd95a4a2b580a0f7873db4e33b8f1cd67a17da91d8765372d1aL91) (not very well though), but it was removed for some reason.

Joe-Edwards-GDS: I suppose the obvious issue is that redeploying the stub will break any active sessions with it, but that hopefully doesn't happen _super_ often.

I'm wondering if there's a lightweight solution - e.g. always using a fixed value. Provided core isn't somehow hardcoding the stub state as well, it should check it can check it's being returned without needing to maintain sessions in the stub.

Joe-Edwards-GDS: If we're doing anything more complex then I'd like to raise a separate ticket, so we can unblock the performance testing (it's been unchecked for ages without causing any problems).

Wynndow (approved):

PR #608, opened 2024-01-16
PYIC-4461 Increase autoscaling maximum

Joe-Edwards-GDS:
Increase the maximum autoscaling limit


di-ipv-orchestrator-stub/deploy/template.yaml
```diff
@@ -481,7 +481,7 @@ Resources:
   OrchStubAutoScalingTarget:
     Type: AWS::ApplicationAutoScaling::ScalableTarget
     Properties:
-      MaxCapacity: 1
+      MaxCapacity: 8
```
Joe-Edwards-GDS, line 484: I think this was overruling the 'desired task count' for the service, so there was still only a single task

TobySaundersGDS (approved):
That makes sense - give this a shot

PR #612, opened 2024-01-18
BAU: Update the orch-stub default vtr value

Wynndow:
### What changed

Update the orch-stub default vtr value

### Why did it change

The vtr value is configurable on the orch-stub. The default is now wrong. The actual orch is now just sending P2. This updates our stub to match.


Joe-Edwards-GDS (approved):
well spotted

PR #613, opened 2024-01-18
BAU: Fix orch-stub vtr parsing

Wynndow:
### What changed

Fix orch-stub vtr parsing

### Why did it change

The value entered in the input on the front page of orch stub was being read as a single array with one string in. This chops the value up on a space delimiter to allow multiple values.

We were also wrapping the array inside a list when setting it on the claimsSet. This just sets the array as the value, which is what we acutally want.

### DIFF
--- a/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
+++ b/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
@@ -48,6 +48,7 @@ import java.net.URISyntaxException;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -86,7 +86,7 @@ public class IpvHandler {
                 String signInJourneyIdText = request.queryMap().get("signInJourneyIdText").value();
                 String[] vtr =
                         request.queryMap().hasKey("vtrText")
-                                ? request.queryMap("vtrText").values()
+                                ? request.queryMap("vtrText").value().split(",")
```
Joe-Edwards-GDS: Do you think it's worth trimming the values as well

Joe-Edwards-GDS (dismissed):

TalatAydinGDS (dismissed):

Joe-Edwards-GDS (approved):

PR #618, opened 2024-01-22
Set tasks to 12 for performance test

TobySaundersGDS:
## Proposed changes
Setting high task count for performance test

### What changed
Set the tasks to 12

### Why did it change
Orch stub is a bottleneck for the IPV Core performance test

This should be turned back down to 1 afterwards to save costs


Joe-Edwards-GDS (approved):
👍  I think this should be ~3x more capacity (although 12x1 CPU vs 1x4)

This is probably not enough to hit full capacity (which is >4x more than the last test), but should be enough to make progress towards it...

PR #647, opened 2024-02-26
PYIC-4373: Upgraded to Node v20

RogVanGDS:
### What changed

Upgraded ticf-stub to Node v20

### Why did it change

As part of our rollout of Node20 now it is a LTS version

### Issue tracking

- [PYI-4373](https://govukverify.atlassian.net/browse/PYI-4373)

### DIFF
--- a/di-ipv-ticf-stub/lambdas/package.json
+++ b/di-ipv-ticf-stub/lambdas/package.json
@@ -9,7 +9,6 @@
     "@aws-lambda-powertools/parameters": "^1.16.0",
     "@aws-sdk/client-dynamodb": "^3.460.0",
     "@aws-sdk/lib-dynamodb": "^3.460.0",
-    "@types/node": "^20.11.20",
     "esbuild": "0.19.8",
     "jose": "^5.1.1",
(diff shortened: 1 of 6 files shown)


di-ipv-ticf-stub/lambdas/package.json
```diff
@@ -9,6 +9,7 @@
     "@aws-lambda-powertools/parameters": "^1.16.0",
     "@aws-sdk/client-dynamodb": "^3.460.0",
     "@aws-sdk/lib-dynamodb": "^3.460.0",
+    "@types/node": "^20.11.20",
```
Joe-Edwards-GDS: Types should be dev-dependencies

(I suspect we also don't actually _need_ this, but it's nice to have it explicit)

Joe-Edwards-GDS (commented):
One tiny comment, but seems reasonable to me.

If you want, you can add an 'engines' restriction in `package.json` to enforce that this is run with node 20, but it's not required (particularly as we're not really relying on any new features).

blakeyp (approved):

PR #650, opened 2024-02-26
BAU: Cache moustache templates

Joe-Edwards-GDS:
The current implementation constructs a new `MoustacheTemplateEngine` for every render, which loads the template from disk every time(!)

The default `MoustacheTemplateEngine` is thread-safe and caches templates by default, so we should use it as such.

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/utils/ViewHelper.java
+++ b/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/utils/ViewHelper.java
@@ -6,7 +6,9 @@ import spark.template.mustache.MustacheTemplateEngine;
 import java.util.Map;
 
 public class ViewHelper {
-    public String render(Map model, String templatePath) {
-        return new MustacheTemplateEngine().render(new ModelAndView(model, templatePath));
+    private final MustacheTemplateEngine templateEngine = new MustacheTemplateEngine();
+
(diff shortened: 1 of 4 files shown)


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/utils/ViewHelper.java
Joe-Edwards-GDS, line 1: This version is non-static and mocked in unit tests...

It would be nice for them all to match, but I wanted to keep the change small.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -225,7 +225,7 @@ private AccessToken exchangeCodeForToken(
             throws OrchestratorStubException, URISyntaxException {
         URI resolve =
                 getIpvBackchannelEndpoint(targetEnvironment).resolve(IPV_BACKCHANNEL_TOKEN_PATH);
-        logger.info("token url is " + resolve);
+        logger.debug("token url is " + resolve);
```
Joe-Edwards-GDS, line 228: (The orch-stub logs are just millions of these messages, obscuring errors and providing minimal value)

Wynndow (approved):

PR #694, opened 2024-03-11
BAU: Refactor to allow error journeys

Joe-Edwards-GDS:
The error journey button doesn't work, as it's missing some required parameters from the main form.

Instead, use a single form for normal/error journeys, using a different formaction to distinguish the two. Also includes some mild refactoring of the handler.


RogVanGDS (dismissed):

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -68,77 +69,96 @@ public class IpvHandler {
...
-                                        getIpvEndpoint(environment).resolve("/oauth2/authorize"))
-                                .requestObject(EncryptedJWT.parse(encryptedJwt.serialize()))
-                                .build();
-
-                response.redirect(authRequest.toURI().toString());
+    public Route doAuthorizeError =
+            (Request request, Response response) -> {
+                var environment = request.queryMap().get(ENVIRONMENT_PARAM).value();
+                var errorType = request.queryMap().get(ERROR_TYPE_PARAM).value();
+                response.cookie("targetEnvironment", environment);
```
Wynndow:
```suggestion
                response.cookie(ENVIRONMENT_PARAM, environment);
```

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
```diff
@@ -68,77 +69,96 @@ public class IpvHandler {
...
-                                .endpointURI(
-                                        getIpvEndpoint(environment).resolve("/oauth2/authorize"))
-                                .requestObject(EncryptedJWT.parse(encryptedJwt.serialize()))
-                                .build();
-
-                response.redirect(authRequest.toURI().toString());
+    public Route doAuthorizeError =
+            (Request request, Response response) -> {
+                var environment = request.queryMap().get(ENVIRONMENT_PARAM).value();
+                var errorType = request.queryMap().get(ERROR_TYPE_PARAM).value();
```
Wynndow, line 102: As you're passing query map in to `getAuthorizeRedirect`, can you extract the error type within that method and save passing an arg?

Joe-Edwards-GDS: No - it's deliberately passed in so that in the non-error case it's ignored (notice how it's always null in the non-error case).

We could pass a boolean flag to control whether it's ignored, but it's much the same effect.

Wynndow: Ah ok, so if error type is not set it doesn't evaluate to null.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/IpvHandler.java
Joe-Edwards-GDS: I think it's technically `ENVIRONMENT_COOKIE` (it just happens to have the same value)

Wynndow (approved):

PR #701, opened 2024-03-13
PYIC-5401: Ensure CRI stubs have consistent users for passport and dl journeys

Wynndow:
### What changed

EnsureCRI stubs have consistent users for passport and dl journeys.

The first commit is just reordering things. The actual changes are in the second commit.

### Why did it change

[PYIC-5401: Group subjects by CRI](https://github.com/govuk-one-login/ipv-stubs/commit/b5c7cc9fae46904f9828a52863e3d2c35bc82677)

This should make it easier to maintain this list. You'll generally want
to add a new subject to a particular CRI, and this makes it easier to
find where to do that. No functional changes in the data.

[PYIC-5401: Add users to passport and DL CRI](https://github.com/govuk-one-login/ipv-stubs/commit/63d93183a71a3eab513ee529a1072970ce35ed5e)

This adds Kenneth as a driving licence CRI user, and Alice as a passport
user.

This will allow the new mitigation routes to be run with a consistent
user. It is possible to run the journeys without this, as we don't do
correlation with failing VCs, it will make for a less confusing
experience.

### Issue tracking

- [PYI-5401](https://govukverify.atlassian.net/browse/PYI-5401)

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/resources/data/criStubData.json
+++ b/di-ipv-credential-issuer-stub/src/main/resources/data/criStubData.json
@@ -829,6 +829,40 @@
       }
     },
     {
+      "criType": "Driving Licence (Stub)",
+      "label": "Kenneth Decerqueira (Valid) DVLA Licence",
+      "payload": {
+        "name": [
+          {
+            "nameParts": [
+              {
+                "value": "Kenneth",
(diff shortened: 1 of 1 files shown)


Joe-Edwards-GDS (approved):
Looks plausible to me

PR #749, opened 2024-04-18
OJ-2454 update with evidence flag for nino

jkunle:
## Proposed changes

image

To

image

### What changed
https://govukverify.atlassian.net/browse/OJ-2454

Scope claim has been removed and an option evidence claim is available for Check HMRC when the
HMRC Check CRI - Evidence Requested 
A new form to specify evidence request setting is presented which defaults to the expected values for NINO

image

To simulate a malformed NINO check set Scoring Policy to `Not gpg45`

image

### Why did it change

In order to influence the behaviour of Check HMRC CRI Credential Issuer the evidence request claim can be passed as part of the JAR, Check Hmrc CRI will use this to determine the nature of Verifiable Credential that is generated


surajkumar:
Hi @jkunle this will affect the other check environments right?

jkunle:
Yeah, it will

surajkumar (approved):

PR #768, opened 2024-04-29
PYIC-3757: Update NINO CRI Stub evidence CI data

shivanshuit914:
## Proposed changes

- Update NINO CRI Stub evidence CI data with correct value

### What changed

- Updated NINO CRI Stub evidence CI to D02

### Why did it change

- To match correct response

### Issue tracking

- [PYI-3757](https://govukverify.atlassian.net/browse/PYI-3757)

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
+++ b/di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
@@ -443,7 +443,7 @@
         "type": "IdentityCheck",
         "strengthScore": 2,
         "validityScore": 0,
-        "ci": ["QQQ"],
+        "ci": ["D02"],
         "failedCheckDetails": [
           {
             "checkMethod": "data"


MikeCollingwood (approved):
I see we have the code used above as well, so happy to approve

di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
```diff
@@ -443,7 +443,7 @@
         "type": "IdentityCheck",
         "strengthScore": 2,
         "validityScore": 0,
-        "ci": ["QQQ"],
+        "ci": ["D02"],
```
DanCorderIPV, line 446: There are two codes in the ticket, do we need a payload for each one, or both in one, or neither?

MikeCollingwood: It's optional between both of them, so if anything it would be to have a second payload with the other code

Though I think example payloads are ok to not be extensive to cover all cases

blakeyp: It's fine just to have the one as an example (just to correct the QQQ placeholder)

PR #770, opened 2024-04-30
[QA-465] : Scale down orch stub

abhinaysrivastavaperf:
There are no performance tests planned for IPV core this week so orch stub is being scaled down for now and will be scaled back up before the next test.


Joe-Edwards-GDS (approved):
Looks like you've reformatted this file, which adds a bunch of noise - can we try and avoid that in future?

PR #771, opened 2024-04-30
PYIC-5954 Created base project module for new EVCS stub. 

rpayal:
## Proposed changes
Create base project module for new EVCS stub. 

### What changed
Created base project module for new EVCS stub. 
Added a initial route to include POST user VCs.

### Issue tracking

- [PYI-5954](https://govukverify.atlassian.net/browse/PYI-5954)

### Other considerations


di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (!userId) {
+    return buildApiResponse({ errorMessage: "Missing userId." }, 400);
+  }
+
+  try {
+    let request;
+    let res: ServiceResponse = {
+      response: undefined
+    };
+    switch(event.httpMethod) {
```
Joe-Edwards-GDS: You might want to consider having separate handlers for each endpoint - that way you don't need this routing logic, it'll all be handled by API Gateway

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (error: any) {
+          console.error(error);
+          return buildApiResponse({ errorMessage: error.message }, 400);
+        }
+        // eslint-disable-next-line no-case-declarations
+        res = await processPostUserVCsRequest(decodeURIComponent(userId), request);
+        break;
+      case "GET":
+        // eslint-disable-next-line no-case-declarations
```
Joe-Edwards-GDS: If you're getting this warning, it probably means you should pull this out into a method and `return` the result.

di-ipv-evcs-stub/lambdas/src/domain/apiResponse.ts
Joe-Edwards-GDS, line 1: Is this used?

di-ipv-evcs-stub/lambdas/src/domain/enums/allState.ts
Joe-Edwards-GDS, line 1: I don't think `All` is necessary here - if you don't think `State` is clear enough then I'd call it `VcState`

di-ipv-evcs-stub/lambdas/src/domain/enums/createState.ts
Joe-Edwards-GDS, line 1: I don't think this is a separate enum - it's just a subset of states that you want to allow on a create request.

Typescript doesn't do any validation, so if you want to make sure the state is the right type, you need to do it explicitly - e.g. when creating new VCs you would need to explicitly check that it's one of the supported 'create' states, and when updating VCs you need to check the target state.

di-ipv-evcs-stub/lambdas/src/domain/enums/vcProvenance.ts
Joe-Edwards-GDS, line 1: Can we use `VcProvenance` etc., I think it's a bit more consistent

di-ipv-evcs-stub/lambdas/src/domain/postRequest.ts
```diff
@@ -0,0 +1,5 @@
+import PersistVC from "./persistVC";
+
+export default interface PostRequest {
+  persistVCs: PersistVC[];
```
Joe-Edwards-GDS: I don't think this is right - it's not a property of the request, it's the request itself.

I think this should be `type PostRequest = PersistVc[];`, or it could just be inlined:
```
export default type PostRequest = {
  vc: string;
  state: VcState;
  metadata?: unknown;
  provenance?: VcProvenance;
}[];
```

di-ipv-evcs-stub/lambdas/src/domain/serviceResponse.ts
Joe-Edwards-GDS, line 1: (Needs updating?)

di-ipv-evcs-stub/lambdas/src/domain/vcMetadata.ts
Joe-Edwards-GDS, line 1: I don't think this metadata interface is fixed - it's an arbitrary json object (this format is just the example that they give)

di-ipv-evcs-stub/lambdas/src/domain/vcStateMetadata.ts
Joe-Edwards-GDS, line 1: A better name might be something like `VcResponseItem`?

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,97 @@
...
+      endpoint: config.localDynamoDbEndpoint,
+      region: config.region,
+    })
+  : new DynamoDB({ region: config.region });
+
+export async function processPostUserVCsRequest(
+    userId: string,
+    postRequest: PostRequest
+  ): Promise<ServiceResponse> {
+  console.info(`Post user record.`);
```
Joe-Edwards-GDS: Is this useful? It should probably either include some useful information, or be removed

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,97 @@
...
+  for (const persistVC of postRequest.persistVCs) {
+      const vcItem: EvcsVcItem = {
+          userId: userId,
+          vc: persistVC.vc,
+          state: persistVC.state,
+          metadata: persistVC.metadata!,
+          provenance: persistVC.provenance!,
```
Joe-Edwards-GDS: Why are we asserting these exist - they should either get a default value or be optional in the `EvcsItem` type

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,97 @@
...
+  userId: string
+): Promise<ServiceResponse> {
+  console.info(`Get user record.`);
+  const getItemInput: GetItemInput = {
+    TableName: config.evcsStubUserVCsTableName,
+    Key: marshall({
+      userId: userId,
+    }),
+  };
+  const items = (await dynamoClient.scan(getItemInput)).Items;
```
Joe-Edwards-GDS: This should be a query not a scan, I think?

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,97 @@
...
+  console.info(`Total VCs retrived - ${JSON.stringify(items?.length)}`);
```
Joe-Edwards-GDS: You shouldn't need to stringify the length, perhaps:
```console.info(`Total VCs retrieved - ${items?.length ?? 0}`);```

File: di-ipv-evcs-stub/lambdas/src/services/evcsService.ts [hunk 8]
```diff
@@ -0,0 +1,97 @@
...
+  const evcsVcItems: EvcsVcItem[] = [];
+  items?.forEach((item) => {
+    evcsVcItems.push(unmarshall(item) as EvcsVcItem);
+  });
```
Joe-Edwards-GDS: You can map here:
```
const items = (await dynamoClient.scan(getItemInput)).Items ?? [];

const evcsVcItems = items.map((item) => unmarshall(item) as EvcsVcItem);
```
(note that this does not actually check that `item` is the right type, but given we own this DB it should be ok)

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  const vcItems: Array<VcStateMetadata> = [];
+  for (const item of evcsVcItems) {
+    const vcItem: VcStateMetadata = {
+        vc: item.vc,
+        state: AllState[item.state as keyof typeof AllState],
+        metadata: item.metadata!,
+    };
+    vcItems.push(vcItem);
```
Joe-Edwards-GDS: and here, chaining together:
```
  const vcs = items
    .map((item) => unmarshall(item) as EvcsVcItem)
    .map((evcsItem) => ({
      vc: evcsItem.vc,
      state: evcsItem.state as VcState,
      metadata: evcsItem.metadata,
    }));
```

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,97 @@
...
+  : new DynamoDB({ region: config.region });
+
+export async function processPostUserVCsRequest(
+    userId: string,
+    postRequest: PostRequest
+  ): Promise<ServiceResponse> {
+  console.info(`Post user record.`);
+  for (const persistVC of postRequest.persistVCs) {
+      const vcItem: EvcsVcItem = {
+          userId: userId,
```
Joe-Edwards-GDS: We seem to be quite inconsistent on indenting - most is 2-space, but occasionally a 4-space creeps in.

Is eslint + prettier configured correctly?

Joe-Edwards-GDS (commented):
I think we're on the right track, but obviously a lot more to do.

The key high-level things to decide are maybe
- whether we want a separate lambda per endpoint, or one shared handler
- how much request validation to perform (by default, TS won't do any explicit validation, it'll just break if you rely on something)

Chris can probably help advise on structure, and I think we've got a few people on the team with good TypeScript experience (e.g. Roger)

I've left a bunch of minor comments as well, but trying not to go into too much detail at this stage.

di-ipv-evcs-stub/lambdas/src/domain/enums/vcProvenance.ts
Joe-Edwards-GDS: Oh I see, these match the names in the swagger documentation. That's probably fine then.

di-ipv-evcs-stub/lambdas/src/domain/enums/allState.ts
rpayal: yeah I also thought so...having one enum for all states.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: Yeah I also thought so but then realise that in CF config will need to have three diff fn's to handle these 3 handlers.

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/domain/enums/createState.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/deploy/template.yaml
```diff
@@ -0,0 +1,408 @@
...
+      - UsePermissionsBoundary
+      - !Ref PermissionsBoundary
+      - !Ref AWS::NoValue
+    CodeSigningConfigArn: !If
+      - UseCodeSigning
+      - !Ref CodeSigningConfigArn
+      - !Ref AWS::NoValue
+    Layers:
+      - !Sub
+        - '{{resolve:secretsmanager:${SecretArn}:SecretString:JAVA_LAYER}}' # or NODEJS_LAYER or PYTHON_LAYER  #pragma: allowlist secret
```
Wynndow: Suspect this should be `NODEJS_LAYER`?

di-ipv-evcs-stub/deploy/template.yaml
```diff
@@ -0,0 +1,408 @@
...
+      FunctionName: !Ref EvcsUserVCsFunction
+      Action: "lambda:InvokeFunction"
+      Principal: !FindInMap [ CoreAccounts, dev02, accountId ]
+
+  EvcsUserVCsFunctionCoreBuildInvokePermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !Ref EvcsUserVCsFunction
+      Action: "lambda:InvokeFunction"
+      Principal: !FindInMap [ CoreAccounts, build, accountId ]
```
Wynndow: Do we need these permissions? The stub is going to sit behind an API gateway, so we won't be invoking anything directly

di-ipv-evcs-stub/lambdas/src/domain/enums/vcState.ts
```diff
@@ -0,0 +1,26 @@
...
+  [VcState.VERIFICATION]: VcState.VERIFICATION,
+} as const;
+
+export const UpdateVcStates = {
+  [VcState.CURRENT]: VcState.CURRENT,
+  [VcState.ABANDONED]: VcState.ABANDONED,
+  [VcState.PENDING_RETURN]: VcState.PENDING_RETURN,
+  [VcState.HISTORIC]: VcState.HISTORIC,
+  [VcState.VERIFICATION_ARCHIVED]: VcState.VERIFICATION_ARCHIVED,
+} as const;
```
Wynndow, line 24: Is there a reason these are defined as objects rather than just arrays?

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+  try {
+    let request;
+    let res: ServiceResponse = {
+      response: Object,
+    };
+    switch (event.httpMethod) {
+      case "POST":
+        try {
+          request = parsePostRequest(event);
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Wynndow: I found this blog post useful when looking at tihs: https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript

di-ipv-evcs-stub/deploy/template.yaml
```diff
@@ -0,0 +1,408 @@
...
+          Properties:
+            RestApiId: !Ref RestApiGateway
+            Path: /vcs/{userId}
+            Method: PATCH
+        GetUserVCs:
+          Type: Api
+          Properties:
+            RestApiId: !Ref RestApiGateway
+            Path: /vcs/{userId}
+            Method: GET
```
Wynndow: I definitley think it'd be worth splitting these out into separate functions. You can still use the same code base, just have a different method as an entrypoint for each lambda.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+        } catch (error: any) {
+          console.error(error);
+          return buildApiResponse({ errorMessage: error.message }, 400);
+        }
+        res = await processPostUserVCsRequest(
+          decodeURIComponent(userId),
+          request,
+        );
+        break;
```
Wynndow: Could inline these to tighten it up a bit
```suggestion
    try {
          res = await processPostUserVCsRequest(
              decodeURIComponent(userId),
              parsePostRequest(event))
          // eslint-disable-next-line @typescript-eslint/no-explicit-any
        } catch (error: any) {
          console.error(error);
          return buildApiResponse({ errorMessage: error.message }, 400);
        }
        break;
```

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+          console.error(error);
+          return buildApiResponse({ errorMessage: error.message }, 400);
+        }
+        res = await processPatchUserVCsRequest(
+          decodeURIComponent(userId),
+          request,
+        );
+        break;
+      case "GET": {
+        let accessTokenVarified;
```
Wynndow:
```suggestion
        let accessTokenVerified;
```

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+            validateAccessToken(event.headers?.Authorisation),
+          );
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (error: any) {
+          console.error(error);
+          return buildApiResponse({ errorMessage: error.message }, 400);
+        }
+
+        if (accessTokenVarified)
+          res = await processGetUserVCsRequest(decodeURIComponent(userId));
```
Wynndow: Do we need to handle the case where the token isn't verified?

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,158 @@
...
+    ExpressionAttributeNames: {
+      "#userId": "userId",
+      "#state": "state",
+    },
+    ExpressionAttributeValues: {
+      ":userIdValue": { S: userId },
+      ":stateValue": { S: VcState.CURRENT },
+    },
+  };
+  // console.info(`Query Input - ${JSON.stringify(getItemInput)}`);
```
Wynndow: Can this be dropped?

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,158 @@
...
+    const vcItem: EvcsVcItem = {
+      userId: userId,
+      vc: persistVC.vc,
+      vcSignature: persistVC.vc.split(".")[2],
+      state: persistVC.state,
+      metadata: persistVC.metadata!,
+      provenance: persistVC.provenance!,
+      ttl: await getTtl(),
+    };
+    await saveUserVC(vcItem);
```
RogVanGDS: Might be worth making this a [Promise All syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) so that we are processing all at the same time, instead of doing one-by-one for performance reasons

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,158 @@
...
+  return {
+    response: {
+      messageId: uuid(),
+    },
+    statusCode: 204,
+  };
+}
+
+async function saveUserVC(evcsVcItem: EvcsVcItem) {
+  console.info(`Save user vc.`);
```
RogVanGDS: Do we have anything like session ID that we can use for all of these logs to give them a bit more context? Might be a bit too convoluted to implement if we don't already have it, but it would definitely make these easier to follow

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,158 @@
...
+  const putItemInput: PutItemInput = {
+    TableName: config.evcsStubUserVCsTableName,
+    Item: marshall(evcsVcItem, {
+      removeUndefinedValues: true,
+    }),
+  };
+  await dynamoClient.putItem(putItemInput);
```
RogVanGDS: Would it be useful to have error handling for these dynamo CRUD interactions? Not sure how likely it will fail but sounds like the kind of thing that would have pretty bad consequences for the service if we silently fail to save user's details

RogVanGDS: At the very least I think we should be logging errors that we can trace / monitor

di-ipv-evcs-stub/lambdas/src/domain/enums/vcState.ts
```diff
@@ -0,0 +1,26 @@
...
+  HISTORIC = "HISTORIC",
+  VERIFICATION_ARCHIVED = "VERIFICATION_ARCHIVED",
+}
+
+export const CreateVcStates = {
+  [VcState.CURRENT]: VcState.CURRENT,
+  [VcState.PENDING]: VcState.PENDING,
+  [VcState.PENDING_RETURN]: VcState.PENDING_RETURN,
+  [VcState.VERIFICATION]: VcState.VERIFICATION,
+} as const;
```
RogVanGDS, line 16: Can this not be something like:
```suggestion
export const isValidCreateVcState = (stateToCheck) => stateToCheck in VcState;
```
Or do we need to construct an object in order to do this? (genuine question, I could be mistaken)

RogVanGDS: I believe when you compile the enum in typescript it already creates a JS object representation of the enum that looks like

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+  console.info(`---Request parsing----`);
+  if (!event.body) {
+    throw new Error("Missing request body");
+  }
+
+  const patchRequest = JSON.parse(event.body);
+  if (
+    !patchRequest ||
+    !patchRequest.updateVCs ||
+    patchRequest.updateVCs.length <= 0 ||
```
RogVanGDS: Can these be collapsed into one? 

Something like `if (!patchRequest?.updateVCs?.length <= 0){...}` then usually it will return undefined if any of the lookups return null instead of giving an NPE

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,158 @@
...
+  console.info(`Patch user record.`);
+  for (const updateVC of postRequest.updateVCs) {
+    const vcItem: EvcsVcItem = {
+      userId: userId,
+      vcSignature: updateVC.signature,
+      state: updateVC.state,
+      metadata: updateVC.metadata!,
+      ttl: await getTtl(),
+    };
+    await updateUserVC(vcItem);
```
RogVanGDS: Same here, might be better to use `Promise.all()`

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
RogVanGDS: EDIT: I can see you have a catch block in the handler, but it doesn't look like we are explicitly checking the status of the update / patch / post requests we make to the database to make sure we are getting a 200 successful response. 

Does this get handled automatically by the DynamoDb client? If so just ignore me 😅

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+  const parts = authheader!.split(" ");
+  if (parts.length != 2) {
+    throw new Error("Invalid access token value");
+  }
+  if (parts[0] != "Bearer") {
+    throw new Error("Access token type must be Bearer");
+  }
+  if (parts[1] === undefined || parts[1] === "") {
+    throw new Error("The access token value must not be null or empty string");
+  }
```
RogVanGDS: Could this actually be null? We are technically only checking against undefined or empty string, so might be worth updating the error message to be accurate

di-ipv-evcs-stub/lambdas/src/domain/vcsResponse.ts
```diff
@@ -0,0 +1,6 @@
+import VcStateMetadata from "./vcStateMetadata";
+
+export default interface VcsResponse {
+  vcs: VcStateMetadata[];
+  afterKey: string;
+}
```
RogVanGDS: Looks like this currently isn't being used anywhere?

di-ipv-evcs-stub/lambdas/src/common/apiResponses.ts
```diff
@@ -0,0 +1,14 @@
+import { APIGatewayProxyResultV2 } from "aws-lambda";
+
+export function buildApiResponse(
```
RogVanGDS, line 3: This can be a default export

di-ipv-evcs-stub/lambdas/src/domain/enums/vcState.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const UpdateVcStates = {
+  [VcState.CURRENT]: VcState.CURRENT,
+  [VcState.ABANDONED]: VcState.ABANDONED,
+  [VcState.PENDING_RETURN]: VcState.PENDING_RETURN,
+  [VcState.HISTORIC]: VcState.HISTORIC,
+  [VcState.VERIFICATION_ARCHIVED]: VcState.VERIFICATION_ARCHIVED,
+} as const;
+
+export default VcState;
```
RogVanGDS: Just a personal preference, but I've always stayed away from default exports in a file where you are also exporting other functions, just to avoid confusion (either that or have them in separate files and default export them)

di-ipv-evcs-stub/deploy/template.yaml
rpayal: done- separation

di-ipv-evcs-stub/deploy/template.yaml
rpayal: yeah seems not needed.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: the outer try...catch will take carte of that ...to return 500 for any failure frm service side.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/domain/enums/vcState.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/domain/vcsResponse.ts
rpayal: removed

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: done

.github/workflows/cri-evcs-stub.yml
```diff
@@ -0,0 +1,75 @@
+name: EVCS Stub Build Build & Push
```
Wynndow: Might be worth renaming this file to just `evcs-stub.yml`, as it's not a CRI

.github/workflows/cri-evcs-stub.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: SAM Validate
+        working-directory: ./di-ipv-evcs-stub/deploy
+        run: sam validate --region ${{ env.AWS_REGION }}
+
+      #- name: SAM Package
+      #  working-directory: ./update-ecs-env-vars/deploy
+      #  env:
+      #    ARTIFACT_BUCKET: ${{ secrets.UPDATE_ECS_ARTIFACT_BUCKET_NAME }}
+      #  run: sam package --s3-bucket="$ARTIFACT_BUCKET" --output-template-file=cf-template.yaml
```
Wynndow: Can this go?

.github/workflows/cri-evcs-stub.yml
rpayal: make sense - done

.github/workflows/cri-evcs-stub.yml
rpayal: done

di-ipv-evcs-stub/deploy/template.yaml
rpayal: ref- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html
So here we r trying to retrieve value of key - JAVA_LAYER    
As our is nodejs proj can try with key- NODEJS_LAYER used only in higher env's

di-ipv-evcs-stub/lambdas/src/domain/enums/vcState.ts
rpayal: Can't use export const isValidCreateVcState = (stateToCheck) => stateToCheck in VcState; 
as the const declared are like sub VcState enum .....contain VcState's valid for create or update.

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: Only thing we got is userId .....which we can't add in log.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: refactored. ----will like  
patchRequest?.updateVCs?.length <= 0

di-ipv-evcs-stub/deploy/template.yaml
Wynndow: Yeah - so this needs updating right to use NODEJS_LAYER rather than JAVA_LAYER?

rpayal: Its  (JAVA_LAYER) a secret key ...which seems sre can confirm if NODEJS_LAYER is configured.

Will update it to NODEJS_LAYER and can try in build.

rpayal: Done - will try change in build.

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done now will be handled by try catch for Promise.all

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: seems its not needed as failure will be handled by validation code.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: implemented a util.

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done default to "ONLINE"

di-ipv-evcs-stub/lambdas/src/domain/vcStateMetadata.ts
rpayal: NA any more

di-ipv-evcs-stub/lambdas/src/domain/postRequest.ts
rpayal: Done

di-ipv-evcs-stub/lambdas/src/domain/vcMetadata.ts
rpayal: Refactor to have random json object instead of fixed vc metadata object.

di-ipv-evcs-stub/lambdas/src/domain/vcMetadata.ts
```diff
@@ -0,0 +1,5 @@
+export default interface VCMetadata {
```
Wynndow: Can this be dropped now that we're treating the incoming metadata as just an object?

I see that it's still being used in the `evcsVcItem`

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -0,0 +1,195 @@
...
+    console.info(`Post user record.`);
+    const allPromises: Promise<PutItemCommandOutput>[] = [];
+    const ttl = await getTtl();
+    for (const postRequestItem of postRequest) {
+      const vcItem: EvcsVcItem = {
+        userId,
+        vc: postRequestItem.vc,
+        vcSignature: postRequestItem.vc.split(".")[2],
+        state: postRequestItem.state,
+        metadata: postRequestItem.metadata!,
```
Wynndow: In the `PostRequest` interface `metadata` is marked as optional, but here we're asserting it's not null. If it ever is I think we'd get a runtime error. Should it have a default of empty object instead?

Wynndow (commented):
Just a couple of small comments, then I think we can get this in

di-ipv-evcs-stub/lambdas/src/domain/vcMetadata.ts
rpayal: done

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
rpayal: done

Wynndow (approved):
Let's give it a go!

PR #782, opened 2024-05-07
LIME-1121 Adding NINO to users and claimset

gds-adhenderson:
### What changed

The Edit User/ New user page has been updated to include a field for manually entering National Insurance Number when accessed via HMRC KBV button. It should be hidden in all other CRIs.

### Why did it change

On the full journey, a user will provide their National Insurance Number in the NINO CRI, before they reach HMRC KBV CRI. In order to simulate this, a NINO field has been added to manually edit a user for this CRI, instead of providing all users with the value from the beginning of the CRI journey. This means a NINO will always need to be added for the CRI to work correctly.

### Issue tracking

- [LIME-1121](https://govukverify.atlassian.net/browse/LIME-1121)

[LIME-1121]: https://govukverify.atlassian.net/browse/LIME-1121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
@@ -70,6 +70,7 @@ public class IdentityMapper {
         FindDateOfBirth dateOfBirth = new FindDateOfBirth(dob, dob);
 
         FullName name = new FullName(map.get("name"), map.get("surname"));
+        String nino = null;
 
         return new Identity(
                 rowNumber,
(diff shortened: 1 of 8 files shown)


di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/config/uatuser/IdentityMapper.java
```diff
@@ -70,6 +70,7 @@ public Identity map(Map<String, String> map, int rowNumber) {
         FindDateOfBirth dateOfBirth = new FindDateOfBirth(dob, dob);
 
         FullName name = new FullName(map.get("name"), map.get("surname"));
+        String nino = "";
```
gds-jbentham: Should this be null rather than empty string

gds-jbentham (approved):

PR #818, opened 2024-05-28
PYIC-6069: Use reverification endpoint

MikeCollingwood:
### What changed

- Change to use reverification endpoint instead of build user identity through user-details endpoint

### Why did it change

- We use the reverification endpoint for mfa resets

### Issue tracking

- [PYI-6069](https://govukverify.atlassian.net/browse/PYI-6069)

### Testing

I will test this:
- manually
- against the e2e regression pack

### DIFF
--- a/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
+++ b/di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
@@ -9,8 +9,6 @@ public class OrchestratorConfig {
                     "IPV_BACKCHANNEL_ENDPOINT", "https://api.identity.build.account.gov.uk/");
     public static final String IPV_BACKCHANNEL_TOKEN_PATH =
             getConfigValue("IPV_BACKCHANNEL_TOKEN_PATH", "/dev/token");
-    public static final String IPV_BACKCHANNEL_USER_IDENTITY_PATH =
-            getConfigValue("IPV_BACKCHANNEL_USER_IDENTITY_PATH", "/dev/user-identity");
     public static final String AUTH_CLIENT_ID = getConfigValue("AUTH_CLIENT_ID", "stubAuth");
     public static final String ORCHESTRATOR_CLIENT_ID =
             getConfigValue("ORCHESTRATOR_CLIENT_ID", "orchestrator");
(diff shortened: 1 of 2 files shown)


di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
```diff
@@ -9,8 +9,6 @@ public class OrchestratorConfig {
                     "IPV_BACKCHANNEL_ENDPOINT", "https://api.identity.build.account.gov.uk/");
     public static final String IPV_BACKCHANNEL_TOKEN_PATH =
             getConfigValue("IPV_BACKCHANNEL_TOKEN_PATH", "/dev/token");
-    public static final String IPV_BACKCHANNEL_USER_IDENTITY_PATH =
-            getConfigValue("IPV_BACKCHANNEL_USER_IDENTITY_PATH", "/dev/user-identity");
```
MikeCollingwood, line 13: The removal of this config from common-infra repos is addressed in this ticket: https://govukverify.atlassian.net/browse/PYIC-6568

Joe-Edwards-GDS, line 13: Do you think it would be preferable to add a new config value for `IPV_BACKCHANNEL_REVERIFICATION_PATH`?

Is there any chance we'll need different paths in different environments? I'm curious about the `/dev/` prefix in the default values, which you've lost.

MikeCollingwood: We currently set `IPV_BACKCHANNEL_USER_IDENTITY_PATH` as "user-details" in each env, and I think allowing different paths per env is unnecessary for a stub since `IPV_BACKCHANNEL_ENDPOINT` distinguishes the endpoints between envs. 

The `dev` config (in core-common-infra) used to have `IPV_BACKCHANNEL_ENDPOINT` be "https://api.identity.build.account.gov.uk/", as it is for `build`. This would mean they'd have to specify different paths to not overlap. Now, it's set as "https://api.cf-dev.dev.identity.account.gov.uk/" (Vam made the change in June 2023)

Implementation is also a bit complicated since each of these config values needs to be set in `stub-common-infra` (`build`) and `core-common-infra` (`dev01` & `02`).

Wynndow (dismissed):
I think this is good, will defer to Joe though on if he wants to keep the paths as config.

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
Joe-Edwards-GDS: Ok sure, happy for us to bake in those paths. (You could do the same with token as well for consistency, but not too fussed)

That dev domain appears broken, so I think it must work differently in dev anyway, but we can worry about that another day.

Wynndow (approved):

PR #823, opened 2024-05-30
BAU: Refactor TS stub unit tests

Joe-Edwards-GDS:
Refactoring tests to run without any AWS dependencies and as part of the pre-merge and post-merge pipelines.


.github/workflows/pre-merge-checks.yml
```diff
@@ -24,3 +24,35 @@ jobs:
           cache: gradle
       - name: Build
         run: ./gradlew clean build
+  build-evcs-stub:
```
Wynndow, line 27: Worth renaming the job on line 12 to be more descriptive, like these?

.github/workflows/pre-merge-checks.yml
```diff
@@ -24,3 +24,35 @@ jobs:
...
+      - name: Install
+        working-directory: ./di-ipv-evcs-stub/lambdas
+        run: npm ci
+      - name: Build
+        working-directory: ./di-ipv-evcs-stub/lambdas
+        run: npm run build
+      - name: Test
+        working-directory: ./di-ipv-evcs-stub/lambdas
+        run: npm run test
+  build-ticf-stub:
```
Wynndow, line 43: I'm wondering if it's worth splitting these out - seems like if you raise a PR against the evcs stub, for example, you'll end up building all the java projects too. I'm not sure if/how we pay for GHA compute, and it's probably minimal anyway, but worth considering?

Wynndow (approved):
The diff makes it a bit hard to compare these directly, but it looks like it's testing the handlers well enough. Are we now missing a bunch of coverage on the evcs service though?

.github/workflows/pre-merge-checks.yml
Joe-Edwards-GDS: So we also have `pre-merge-ticf.yaml` etc., and exclude those paths from this one?

It might be nice, although I think not critical. GHA compute is an issue - there's a free tier limit for the org, but atm this will be dwarfed by the IPV Core integration test runs 😬

Joe-Edwards-GDS:
> Are we now missing a bunch of coverage on the evcs service though?
@Wynndow yes :( though it's a bit tricky:
- The `evcsService` is highly dependent on the AWS SDK, so will require a somewhat more elaborate mocking configuration to test
- The `evcsService` doesn't _do_ a lot except map onto the SDK methods, so had a bit less value in testing anyway
- I want to get the tests running at all first, then we can worry about coverage!

In an ideal world, I think we'd have a data layer which is a very thin wrapper around the SDK and keep any business logic in a service layer. There's also some logic here that could be pulled up into the handler (e.g. the service has no business defining the response status code). If I do any further refactoring I might try and move us towards that world.

.github/workflows/pre-merge-checks.yml
Wynndow: Yeah. that's what i was thinking. If we go down that path it might be worth splitting apart the java projects too

Joe-Edwards-GDS: I'm going to merge this as-is, and consider doing this as a future enhancement - the volume of PRs in this repo isn't super high anyway :)

PR #842, opened 2024-06-12
PYIC-6782 Update stub to decode state query param value. Also update orch stub to have 'sub' instead of 'subject' in eves token jwt.

rpayal:
## Proposed changes
Update stub to decode state query param value.
Update orch stub to have 'sub' instead of 'subject' in eves token jwt.

### What changed
Change in evcs and orch stub.

### Issue tracking

- [PYI-6782](https://govukverify.atlassian.net/browse/PYI-6782)

### Other considerations

### DIFF
--- a/di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
+++ b/di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
@@ -160,7 +160,9 @@ function isValidUpdateVcState(patchRequest: PatchRequest[]): boolean {
 
 function getRequestedStates(event: APIGatewayProxyEvent): string[] {
   const STATE_ALL = "ALL";
-  const stateInQuery = event.queryStringParameters?.state;
+  const stateInQuery = decodeURIComponent(
+    event.queryStringParameters?.state ?? "",
+  );
   let requestStates = Array.from((stateInQuery || VcState.CURRENT).split(","));
   requestStates.forEach(function (state) {
     if (state != STATE_ALL && !(state in VcState)) {
--- a/di-ipv-evcs-stub/openAPI/evcs-external.yaml
+++ b/di-ipv-evcs-stub/openAPI/evcs-external.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  title: "EVCS Stub External"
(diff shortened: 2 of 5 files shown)


di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
Joe-Edwards-GDS, line 1: I thought you said it already worked without decoding the parameter - is that actually not the case?

(We shouldn't decode it twice - that could lead to other issues, if we ever wanted a value with a genuine `%` in it)

Joe-Edwards-GDS (commented):
Change to the sub looks sensible - but I'm keen to make sure we handle the encoding correctly.

di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
rpayal: Yeah it was.
Tested it with both query param as "state=CURRENT%2CABANDONED" and "state=CURRENT,ABANDONED"

Joe-Edwards-GDS: Ok - if it was working before, then do we still need to change it?

rpayal: Test with "CURRENT%2CPENDING_RETURN" will not work without state value decode.

di-ipv-evcs-stub/openAPI/evcs-external.yaml
```diff
@@ -58,6 +58,7 @@ paths:
           $ref: '#/components/responses/BadRequest'
         500:
           $ref: '#/components/responses/ServerError'
+      x-amazon-apigateway-request-validator: ALL
```
Joe-Edwards-GDS, line 61: (Presumably this means we could remove our in-code validation logic - but no urgency)

Joe-Edwards-GDS (approved):

di-ipv-evcs-stub/openAPI/evcs-external.yaml
rpayal: Sure will test it and remove it from code.

PR #867, opened 2024-06-28
LS-2207 Get JMX metrics published to Cloudwatch Insights

andyloughran:
### What changed

Wraps the mocks in a jmx_exporter class to show metrics on port 9404.  CloudWatch + containerinsights will then ingest those metrics.

### Why did it change

To better understand mock behaviour under load.

### DIFF
--- a/di-ipv-orchestrator-stub/build.gradle
+++ b/di-ipv-orchestrator-stub/build.gradle
@@ -1,4 +1,5 @@
 plugins {
+	id 'com.github.johnrengelman.shadow' version '8.1.1'
 	id 'idea'
 	id 'application'
 	id 'java'
--- a/di-ipv-orchestrator-stub/build.gradle
+++ b/di-ipv-orchestrator-stub/build.gradle
@@ -1,5 +1,4 @@
 plugins {
-	id 'com.github.johnrengelman.shadow' version '8.1.1'
 	id 'idea'
 	id 'application'
 	id 'java'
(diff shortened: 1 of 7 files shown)


di-ipv-orchestrator-stub/build.gradle
```diff
@@ -1,4 +1,5 @@
 plugins {
+	id 'com.github.johnrengelman.shadow' version '8.1.1'
```
andyloughran: This was used to generate a fat jar.  If there's a more appropriate way, please do change this.

Joe-Edwards-GDS (commented):
I've deployed this to my dev environment. The orch stub still appears to run as expected, and I can see the new docker tags/port mappings.

I can't see the metrics in CloudWatch though - do we need to set up the cloudwatch agent? i.e. https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights-Prometheus-install-ECS.html

andyloughran:
Thanks Joe, I'd expected the agent to be present given we're running fargate, so I'll re-review the documentation to see if there are other steps I've missed.  There's no doc explicitly covering this exact use case that I could find.

andyloughran:
> I've deployed this to my dev environment. The orch stub still appears to run as expected, and I can see the new docker tags/port mappings.
> 
> I can't see the metrics in CloudWatch though - do we need to set up the cloudwatch agent? i.e. https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights-Prometheus-install-ECS.html

You're bang on the money.  I hadn't noted that bit.  I think if we are to do that we need to fork the template so we're keeping track of it, and we probably want the version that's the same networking mode as the stub.

PR #890, opened 2024-07-16
PYIC-6021: added private cimit api gateway

sam803:
## Proposed changes

Add a private api gateway for cimit lambdas

### Why did it change

The CIMIT interface is being updated to use a private API Gateway to invoke its services.

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6021


Joe-Edwards-GDS (approved):
Looks plausible to me - pretty much identical to the real CIMIT implementation.

If there are any issues, they'll probably come out when @thebauSoftwire tries to use it :)

PR #919, opened 2024-07-23
BAU: Fix cimit stub logging

Wynndow:
### What changed

Fix cimit stub logging

### Why did it change

A previous commit removed the use of powertools for logging. I wasn't thorough enough though and logging is broken.

The log4j-layout-template-json lambda needs adding - powertools must have been bringing it in.

We also need to remove the powertool resolver from the logging config, otherwise it blows up.


thebauSoftwire (approved):
Thanks for being so quick on this!

PR #924, opened 2024-07-25
PYIC-6021: update cimit stub

thebauSoftwire:
### What changed
- updates response shape for PostMitigations lambda to match real CIMIT endpoint + updates unit tests
- converts Internal api to regional
- applies custom domain to public internal api so it can be called from a VPC

### Why did it change

The CIMIT interface is being updated to use a private API Gateway so the CIMIT stub should be updated to match. Some changes made e.g. making it public and adding a custom domain so it can be easily accessible to other accounts

### Issue tracking

- [PYIC-6021](https://govukverify.atlassian.net/browse/PYIC-6021)

[PYIC-6021]: https://govukverify.atlassian.net/browse/PYIC-6021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-cimit-stub/lambdas/post-mitigations/src/main/java/uk/gov/di/ipv/core/postmitigations/domain/PostMitigationsResponse.java
+++ b/di-ipv-cimit-stub/lambdas/post-mitigations/src/main/java/uk/gov/di/ipv/core/postmitigations/domain/PostMitigationsResponse.java
@@ -1,7 +1,3 @@
 package uk.gov.di.ipv.core.postmitigations.domain;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import lombok.Builder;
-
-@Builder
-public record PostMitigationsResponse(@JsonProperty("result") String result) {}
+public record PostMitigationsResponse(String result) {}
--- a/di-ipv-cimit-stub/lambdas/post-mitigations/src/main/java/uk/gov/di/ipv/core/postmitigations/domain/PostMitigationsResponse.java
+++ b/di-ipv-cimit-stub/lambdas/post-mitigations/src/main/java/uk/gov/di/ipv/core/postmitigations/domain/PostMitigationsResponse.java
@@ -1,0 +1,7 @@
+package uk.gov.di.ipv.core.postmitigations.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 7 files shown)


di-ipv-cimit-stub/lambdas/post-mitigations/src/main/java/uk/gov/di/ipv/core/postmitigations/domain/PostMitigationsResponse.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.di.ipv.core.postmitigations.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+
+@Builder
```
Joe-Edwards-GDS: (not sure you need builder here but doesn't hurt)

di-ipv-cimit-stub/lambdas/post-mitigations/src/main/java/uk/gov/di/ipv/core/postmitigations/domain/PostMitigationsResponse.java
```diff
@@ -0,0 +1,7 @@
...
+public record PostMitigationsResponse(@JsonProperty("result") String result) {}
```
Joe-Edwards-GDS: (you don't generally need `@JsonProperty` if it matches the param name, unless you think it helps clarity)

di-ipv-cimit-stub/openAPI/cimit-internal.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.3
 info:
-  title: CIMITPrivateAPI
+  title: CIMITInternalAPI
   description: API Gateway Experiment
```
Joe-Edwards-GDS: Can we change this from 'Experiment' as well, to be a bit more descriptive...

Joe-Edwards-GDS: Not much more descriptive, but it'll do!

Joe-Edwards-GDS (approved):

PR #933, opened 2024-07-29
BAU: Update orch ports and default config

Joe-Edwards-GDS:
### What changed

- Update the default config for local running
- Update README

### Why did it change

So orch-stub can be run with a one-click run process with core-front and core-back to have a complete local running setup


di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
Joe-Edwards-GDS, line 1: These config defaults are essentially copied from the `compose.yaml` file in core-back local-running, which is already public. If we use a more sophisticated config mechanism, this might change.

Wynndow (approved):

PR #940, opened 2024-08-01
PYIC-7173 Secure queue stub

Joe-Edwards-GDS:
### What changed

- Change queue stub URL to
  - `queue.stubs.account.gov.uk` (stubs-prod)
  - `queue.build.stubs.account.gov.uk` (stubs-build)
- Add API key

### Why did it change

We want to secure the queue API and start using the production version

*N.B. this will need to be coordinated with an update to the F2F stub to provide the API key*


Wynndow (dismissed):
Will this also need a change to the selenium tests

Wynndow (approved):

PR #946, opened 2024-08-07
PYIC-7174 dequeue from queue-stub

Joe-Edwards-GDS:
### What changed

Added dequeue capability to queue-stub

### Why did it change

Local-running cannot consume the SQS queues directly (which would require AWS permissions), so we want to provide an interface to consume them instead.

### Issue tracking

- [PYIC-7174](https://govukverify.atlassian.net/browse/PYIC-7174)

[PYIC-7174]: https://govukverify.atlassian.net/browse/PYIC-7174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-queue-stub/enqueue-event/src/services/queueService.ts
+++ b/di-ipv-queue-stub/enqueue-event/src/services/queueService.ts
@@ -1,0 +1,117 @@
+import {
+  AddPermissionCommand,
+  CreateQueueCommand,
+  DeleteMessageCommand,
+  GetQueueUrlCommand,
+  Message,
+  QueueDoesNotExist,
+  ReceiveMessageCommand,
+  SendMessageCommand,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+
+const sqsClient = new SQSClient({ region: "eu-west-2" });
+
+export const getOrCreateSqsQueueUrl = async (
+  accountId: string,
+  queueName: string,
+): Promise<string> => {
+    try {
+        const queueUrlResponse = await sqsClient.send(new GetQueueUrlCommand({
+            QueueName: queueName,
+        }));
+        console.info(`Queue already exists ${queueName}`);
+        return queueUrlResponse.QueueUrl!;
+    } catch (error) {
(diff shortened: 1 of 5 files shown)


di-ipv-queue-stub/enqueue-event/src/services/queueService.ts
```diff
@@ -0,0 +1,112 @@
...
+            await sqsClient.send(new AddPermissionCommand({
+                QueueUrl: createQueueResponse.QueueUrl,
+                Label: "add read and delete from dev, build, staging, prod perms",
+                AWSAccountIds: [
+                    "457601271792",
+                    "130355686670",
+                    "175872367215",
+                    "335257547869",
+                    "991138514218",
+                    "075701497069"
```
blakeyp: I know this was copied from before but do these perms really need adding for int/prod?

di-ipv-queue-stub/enqueue-event/src/services/queueService.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            console.info("Attached permissions");
+
+            return createQueueResponse.QueueUrl!;
+        }
+        throw error;
+    }
+};
+
+export const enqeueEvent = async (
```
blakeyp: typo *enqueue ?

Joe-Edwards-GDS: gah - why is queueueue so hard to type

di-ipv-queue-stub/enqueue-event/src/services/queueService.ts
Joe-Edwards-GDS: Fair point - doesn't seem likely we'll ever want int/prod to subscribe to stub queue events (pretty unlikely for staging as well tbh, but more plausible)

blakeyp (approved):

PR #956, opened 2024-08-16
PYIC-7140 migrate cri stub to javalin

Joe-Edwards-GDS:
### What changed

Update from Spark Java to Javalin as a web framework

### Why did it change

Spark Java is unsupported and hard to keep patched

### Issue tracking

- [PYIC-7140](https://govukverify.atlassian.net/browse/PYIC-7140)

[PYIC-7140]: https://govukverify.atlassian.net/browse/PYIC-7140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/App.java
```diff
@@ -1,59 +1,7 @@
...
-                                .flatMap(Collection::stream)
-                                .filter(
-                                        t ->
-                                                t.getClass()
-                                                        .isAssignableFrom(
-                                                                HttpConnectionFactory.class))
-                                .map(t -> ((HttpConnectionFactory) t).getHttpConfiguration())
-                                .forEach(
-                                        t -> {
-                                            t.setRequestHeaderSize(9 * 1024);
```
Joe-Edwards-GDS, line 51: This bumped the header size from 8KB to 9KB...

We can do the same, but it feels a bit suspicious - running through a journey in build I can see max ~3KB for the request URL + headers, except for the `referer` header on the form post.

I think we can probably leave this out, and I've added a `meta` directive to the template to strip the path/query from the referrer (otherwise it contains the entire JAR request again).

Wynndow: Sounds like stripping stuff form the referer would help a lot. If we try it without increasing this and things go wrong, and least we'll have an idea of what to look at.

Wynndow (approved):
I can't see anything concerning here - as long as it works lets give it a go

PR #1013, opened 2024-10-03
PYIC-7425: Update management api request body

thebauSoftwire:
### What changed
- txn, ci and type now grouped in `evidence` in request body
- statusCode now accepted in request body
  - removes statusCode from path

### Why did it change
Status codes were inconsistently using a separate path when requesting for ticf to return an error which resulted in messier stub code so we combine it instead in the body of the request

### Issue tracking

- [PYIC-7425](https://govukverify.atlassian.net/browse/PYIC-7425)

[PYIC-7425]: https://govukverify.atlassian.net/browse/PYIC-7425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


di-ipv-ticf-stub/lambdas/jest.config.test.js
```diff
@@ -4,8 +4,6 @@ module.exports = {
   coverageDirectory: "coverage",
   coverageProvider: "v8",
 
-  preset: "@shelf/jest-dynamodb",
-
```
thebauSoftwire, line 8: This was preventing me from running the unit tests locally and I think we mock methods which make calls to dynamodb in the tests anyway so not sure why need this jest preset

Joe-Edwards-GDS: Is this used anywhere now? - Can we remove it from `package.json`

di-ipv-ticf-stub/lambdas/src/handlers/ticfHandler.ts
```diff
@@ -2,33 +2,31 @@ import { APIGatewayProxyEventV2, APIGatewayProxyResultV2 } from "aws-lambda";
...
+      { errorMessage: error.message },
+      error.statusCode || 500,
+    );
   }
 }
 
 function parseRequest(event: APIGatewayProxyEventV2): TicfRequest {
   if (!event.body) {
-    throw new Error("Missing request body");
+    throw new BadRequestError("Missing request body");
```
Joe-Edwards-GDS, line 29: Nice!

di-ipv-ticf-stub/lambdas/src/management/handlers/managementHandler.ts
```diff
@@ -14,51 +14,49 @@ export async function handler(
...
+  const ticfManagementRequest: TicfManagementRequest = JSON.parse(event.body);
 
-    if (ticfManagementRequest && !ticfManagementRequest.type) {
-      throw new Error("Invalid request");
-    }
+  if (
+    !ticfManagementRequest.evidence?.type &&
+    (!ticfManagementRequest.statusCode || ticfManagementRequest.statusCode == 200)
+  ) {
+    throw new BadRequestError("Invalid request - evidence type required if configuring a successful TICF response");
```
Joe-Edwards-GDS: Do we actually need to validate this - what would happen if we didn't?

Fine to leave - I don't think we particularly need to simulate an invalid response with a 200.

di-ipv-ticf-stub/lambdas/src/management/services/userEvidenceService.ts
```diff
@@ -16,15 +17,14 @@ const dynamoClient = config.isLocalDev
...
   ticfManagementRequest: TicfManagementRequest,
-  statusCode: number,
 ): Promise<void> {
-  const { responseDelay, ...ticfEvidenceItem } = ticfManagementRequest;
+  const { responseDelay, statusCode, evidence } = ticfManagementRequest;
 
   const userEvidence: UserEvidenceItem = {
     userId: userId,
-    evidence: ticfEvidenceItem,
+    evidence: evidence || ({} as TicfEvidenceItem),
```
Joe-Edwards-GDS: if this is `{}` will the 'default' logic work when you try to retrieve it? Shouldn't it just stay undefined?

Joe-Edwards-GDS (dismissed):
Thanks for tidying some code at the same time!

Couple of optional comments

di-ipv-ticf-stub/lambdas/src/management/handlers/managementHandler.ts
thebauSoftwire: Just covering edge cases since both `statusCode` and `evidence` are optional but you're right, probably a case that we don't need to verify. I did leave a check for if an evidence is provided, it should contain a `type` to make sure someone can't just do something like `{evidence: "not-the-right-thing"}`

thebauSoftwire: Would be cleaner with a schema validator like yup but seems a bit much for this 😅

Joe-Edwards-GDS: Yeah - you could do some `typeof` checks (to make sure it's an object) but it's only a stub, so let's not worry too much!

Joe-Edwards-GDS (approved):

PR #1015, opened 2024-10-04
OJ-2811: Generate international user context claim for Address CRI

mikebeeby:
### What changed

Added new buttons with label `Address CRI <env> - International Address` with query param `context=international_user` 
Refactored createJWTClaimsSet to add the context to the Auth JAR if present.

### Why did it change

Core stub needs to be updated to generate and pass the claim context: international_user , so that we can correctly simulate and test journeys for users with non-UK addresses.

### Issue tracking

- [OJ-2811](https://govukverify.atlassian.net/browse/OJ-2811)

[OJ-2811]: https://govukverify.atlassian.net/browse/OJ-2811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/handlers/CoreStubHandler.java
+++ b/di-ipv-core-stub/src/main/java/uk/gov/di/ipv/stub/core/handlers/CoreStubHandler.java
@@ -277,12 +277,14 @@ public class CoreStubHandler {
             request.session().removeAttribute("evidence_request");
         }
 
+        var context = request.queryParams("context");
+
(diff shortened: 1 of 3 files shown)


Caitlin-cooling (approved):
Changes look good to me! 

This stub is starting to get a bit out of hand though think we need to look at a redesign to something more like this is the future https://rp-integration.build.stubs.account.gov.uk/ to give us a bit more flexility with the options

PR #1049, opened 2024-10-17
PYIC-7568 Set orch-stub environment variables directly

Joe-Edwards-GDS:
### What changed

Rather than using SSM parameters as the default value for template parameters, read the SSM values at deploy time.

### Why did it change

SSM parameter defaults are calculated at template creation and never updated - this means that changing the default value will not update any stacks that are already deployed.

If we read the value directly, then behaviour is more predictable.

### DIFF
--- a/di-ipv-orchestrator-stub/deploy/template.yaml
+++ b/di-ipv-orchestrator-stub/deploy/template.yaml
@@ -27,118 +27,6 @@ Parameters:
     Description: "The ARN of the permissions boundary to apply when creating IAM roles"
     Type: String
     Default: "none"
-  CodeSigningConfigArn:
-    Type: String
-    Description: Asserts that lambdas are signed when deployed.
-    Default: none
-  DeploymentStrategy:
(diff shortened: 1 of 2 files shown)


di-ipv-orchestrator-stub/deploy/template.yaml
```diff
@@ -27,118 +27,6 @@ Parameters:
...
-    Description: "Predefined deployment configuration for ECS application"
-    Type: String
-    Default: "None"
-    # Allowed values: See https://docs.aws.amazon.com/codedeploy/latest/userguide/deployment-configurations.html
-    AllowedValues:
-      - None
-      - CodeDeployDefault.ECSCanary10Percent5Minutes
-      - CodeDeployDefault.ECSCanary10Percent15Minutes
-      - ECSCanary50Percent5Minutes
-      - CodeDeployDefault.ECSAllAtOnce
```
Joe-Edwards-GDS, line 44: This was used for the canary deployment resources, which have since been removed

Wynndow (approved):

PR #1073, opened 2024-10-29
PYIC-7624: Serve enc key from well-known endpoint

Wynndow:
This configures the private encryption key to be used for decrypting JARs in the CRI stub to be the same for all clients. This means we only need to serve on key on the well-known endpoint.

### What changed

Serve enc key from well-known endpoint

### Why did it change

This configures the private encryption key to be used for decrypting JARs in the CRI stub to be the same for all clients. This means we only need to serve on key on the well-known endpoint.

### Issue tracking

- [PYIC-7624](https://govukverify.atlassian.net/browse/PYIC-7624)

[PYIC-7624]: https://govukverify.atlassian.net/browse/PYIC-7624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (dismissed):
Seems plausible to me, assuming it won't cause orch to fall over.

The VC signing key will be published on a `did.json` endpoint instead, so we can exclude that for now.

Joe-Edwards-GDS:
(once you've fixed the spotless issues I guess!)

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1080, opened 2024-10-30
PYIC-7625: Orch stub jwks endpoint

Wynndow:
### What changed

Orch stub jwks endpoint

### Why did it change

[PYIC-7625: Consume JWKs for signing](https://github.com/govuk-one-login/ipv-stubs/commit/48138815c03f56e44e98ea99d38817641b9c23e3)

This updates the signing logic to consume private keys in JWK format,
rather than in pkcs8. It makes parsing the private key material simpler.

Also tidies up a few bits and bobs, including only creating a JWS signer
once rather than for every signing operation. This could help improve
performance of the stub under load.

[PYIC-7625: Orch stub well-known endpoint](https://github.com/govuk-one-login/ipv-stubs/commit/5722aa936bd1307773e7671c4b96367703fe8b5b)

This adds a new well-known endpoint to serve the public siging key used
by orch to sign JARs.

[PYIC-7625: Include kid in JWS and JWEs](https://github.com/govuk-one-login/ipv-stubs/commit/175159682af990c7787516f4b703edb2b2338433)

This change includes the appropriate kid in the header for signed JWTs
and encrypted JARs.

### Issue tracking

- [PYIC-7625](https://govukverify.atlassian.net/browse/PYIC-7625)

[PYIC-7625]: https://govukverify.atlassian.net/browse/PYIC-7625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/handlers/JwksHandler.java
```diff
@@ -0,0 +1,34 @@
...
+        ctx.json(
+                new JWKSet(
+                                List.of(
+                                        new ECKey.Builder(
+                                                        ECKey.parse(ORCHESTRATOR_SIGNING_JWK)
+                                                                .toPublicJWK())
+                                                .keyID(
+                                                        KeyIdGenerator.generate(
+                                                                ORCHESTRATOR_SIGNING_JWK,
+                                                                ORCHESTRATOR_CLIENT_ID))
```
Wynndow: This will only list a signing key with kid used when the orch stub is operating as itself. We may in future want to include a key with the keyId used when acting as a stub for auth (mfa reset)

Joe-Edwards-GDS: I'm a bit less sure about this, we'll definitely need to support MFA reset... Given that we have a single orch stub, it's perhaps more realistic to just present a single consistent key on its JWKS endpoint? Or set up multiple different endpoints for orch vs. auth?

I'm also leaning towards configuring the `kid` in the JWK config itself - it feels less surprising than ignoring the `kid` there and generating a new one dynamically each time you use it...

Wynndow: Regarding the first point - given that we'll be now including key IDs in the signature headers, can we just have two keys on the same endpoint? Sort of the point, no?

And for the second, yeah i think I agree having mulled on it a it more. I'll remove the logic and just add something to the config. How do you feel about the format though?

Joe-Edwards-GDS: I think the format seems reasonable - i.e. having a human-readable bit to make it easy to identify when copied around, but also a random/hashed part to avoid clashes.

I guess you're right about including multiple keys - for a slightly different reason than prod will be, but I guess it'll exercise the key-selection logic in core-back!

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
```diff
@@ -11,26 +11,31 @@ public class OrchestratorConfig {
...
             getConfigValue("ORCHESTRATOR_REDIRECT_URL", "http://localhost:4500/callback");
-    public static final String ORCHESTRATOR_CLIENT_SIGNING_KEY =
+    public static final String ORCHESTRATOR_SIGNING_JWK =
             getConfigValue(
-                    "ORCHESTRATOR_CLIENT_SIGNING_KEY",
+                    "ORCHESTRATOR_SIGNING_JWK",
                     // This is a test key used for local development
-                    "MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgOXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthWhRANCAAQT1nO46ipxVTilUH2umZPN7OPI49GU6Y8YkcqLxFKUgypUzGbYR2VJGM+QJXk0PI339EyYkt6tjgfS+RcOMQNO"); // pragma: allowlist secret
-    public static final String ORCHESTRATOR_DEFAULT_JAR_ENCRYPTION_PUBLIC_KEY =
+                    "{\"kty\":\"EC\",\"d\":\"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"); // pragma: allowlist secret
```
Joe-Edwards-GDS: Do these default keys need kid/use as well?

Joe-Edwards-GDS (dismissed):
Looks good, but I think we might need to update the default keys?

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
Wynndow: I don't think these are actually ever used. I'll add them in though just in case

Joe-Edwards-GDS (approved):

di-ipv-orchestrator-stub/src/main/java/uk/gov/di/ipv/stub/orc/config/OrchestratorConfig.java
Joe-Edwards-GDS: I was going to say local running, but looks like we set those explicitly as well :shrug:

PR #1098, opened 2024-11-07
PYIC-7631: Remove basic auth from jwks endpoint

Wynndow:
### What changed

Remove basic auth from jwks endpoint

### Why did it change

We don't need it and it makes the config for core more tricky.

There's no `beforeExcept` filter, so this seemed easier than configuring the basic auth handler on each endpoint directly.

### Issue tracking

- [PYIC-7631](https://govukverify.atlassian.net/browse/PYIC-7631)

[PYIC-7631]: https://govukverify.atlassian.net/browse/PYIC-7631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (approved):
I think this makes sense - the only endpoint that _needs_ basic auth is the homepage route, but it feels safer to work by exception.

PR #1189, opened 2025-01-23
BAU: Add International Address stub data

AmritSidhu:
### What changed

Add International Address stub data - so we can run e2e tests for international address journey


di-ipv-credential-issuer-stub/src/main/resources/data/criStubEvidencePayloads.json
```diff
@@ -33,6 +33,19 @@
         "strengthScore": 4
       }
     },
+    {
+      "criType": "Fraud Check (Stub)",
+      "label": "Failed fraud check (M1C)",
```
blakeyp: Could we label this intl' address - as there will be a different VC for M1C for Experian being down

blakeyp (approved):

PR #1391, opened 2025-04-17
PYIC-8202: Update stub data

thebauSoftwire:
### What changed
Consolidating stub data to match across all of our CRI stubs 

### Why did it change
Not all of our data matched between different test users across the different stubs. This PR aims to tidy that up.

### Issue tracking

- [PYIC-8202](https://govukverify.atlassian.net/browse/PYIC-8202)

[PYIC-8202]: https://govukverify.atlassian.net/browse/PYIC-8202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


di-ipv-credential-issuer-stub/src/main/resources/data/criStubData.json
```diff
@@ -292,11 +292,11 @@
           {
             "nameParts": [
               {
-                "value": "Böb",
+                "value": "Bob",
```
thebauSoftwire, line 295: @SreeBandla01 were the use of these characters intentional?

thebauSoftwire:
Probably not part of this ticket but wondering if there would be a more flexible way of creating the data for the stubs that would make them easier to maintain e.g. have a single source of truth for the above users 🤔

AmritSidhu:
If we are keeping this format I have a little suggestion. Could we group the names together per stub? As we've added more data over time, names get added to the bottom of the lists, would be nice to keep them together, for example you can see the kenneths in different order (maybe even in alphabetised order) 
Screenshot 2025-04-17 at 17 12 30

thebauSoftwire:
Thanks :) have collected same users together and alphabetically ordered them

di-ipv-credential-issuer-stub/src/main/resources/data/criStubData.json
SreeBandla01: Not that I aware of and we dont need them afaik

SreeBandla01: If any one needs a specific format of the name for a specific test they can use manually as one off and if above data used in other teams? not sure and if they use in any other tests they might fail, better to confirm from relavent teams

Joe-Edwards-GDS: I expect these were put in to test the name correlation logic, but IMO it'd be better to test that explicitly (and if we do want it to be tested 'implicitly', it should be via API tests not here)

SreeBandla01 (approved):

PR #1421, opened 2025-05-08
PYIC-8285: make 400 error messages more specific and return 400 with …

thebauSoftwire:
…duplicate VCs

### What changed
- making 400 error messages more specific
- returning a 400 with duplicate VCs

### Why did it change
Changes made as part of QA feedback

### Issue tracking

- [PYIC-8285](https://govukverify.atlassian.net/browse/PYIC-8285)

## Checklists

[PYIC-8285]: https://govukverify.atlassian.net/browse/PYIC-8285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
+++ b/di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
@@ -167,25 +167,40 @@ function parsePutRequest(event: APIGatewayProxyEvent): PutRequest {
   const parsedPutRequest = JSON.parse(event.body);
 
   // Validate top-level request attributes
-  if (
-    !parsedPutRequest.userId ||
-    !parsedPutRequest.vcs ||
-    parsedPutRequest.vcs.length === 0
-  ) {
-    throw new Error("Invalid request");
+  if (!parsedPutRequest.userId) {
+    throw new Error("Invalid request: missing userId");
+  }
+
+  if (!parsedPutRequest.vcs || parsedPutRequest.vcs.length === 0) {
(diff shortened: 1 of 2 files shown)


di-ipv-evcs-stub/lambdas/src/handlers/evcsHandler.ts
```diff
@@ -167,25 +167,40 @@ function parsePutRequest(event: APIGatewayProxyEvent): PutRequest {
...
     }
   });
 
+  const uniqueVcs = new Set(
+    parsedPutRequest.vcs.map((vc: VcDetails) => vc.vc),
+  );
+  if (uniqueVcs.size !== parsedPutRequest.vcs.length) {
+    throw new Error("Invalid vcs: cannot have duplicate VCs in request");
+  }
+
```
thebauSoftwire, line 195: @blakeyp @aleemriaz I'm throwing a 400 here for any duplicate VCs, not just if they have the same state. This differs from the AC which says that we should throw if we have duplicate VCs with the same states. 

I don't think it should be just when they have the same state as EVCS would have to know which of the two VCs to store. I also don't think this should ever happen in practice anyway but let me know if you disagree

aleemriaz: Agreed, any duplicate VCs, regardless of state should result in a 400.

aleemriaz (approved):
Approved

PR #1436, opened 2025-05-14
BAU: Add orch stub perf option

MikeCollingwood:
### What changed

- Add orch stub perf option

### Why did it change

- The perf orch stub caused the majority of the 5xx issues and was the leading cause of latency. In order to make it behave closer to build, this allows us to additionally trigger a performance journey from the main orch stub.

## Checklists


AmritSidhu (approved):

di-ipv-orchestrator-stub/core-dev-deploy/template.yaml
```diff
@@ -466,6 +466,8 @@ Resources:
               Value: '{{resolve:ssm:/stubs/orch/env/AUTH_SIGNING_JWK}}'
             - Name: ORCHESTRATOR_DEFAULT_JAR_ENCRYPTION_PUBLIC_JWK
               Value: '{{resolve:ssm:/stubs/orch/env/ORCHESTRATOR_DEFAULT_JAR_ENCRYPTION_PUBLIC_JWK}}'
+            - Name: ORCHESTRATOR_DEV_JAR_ENCRYPTION_PUBLIC_JWK
```
AmritSidhu, line 469: Will these values need to be added to the core.dev01/02.params.yaml because I know the surrounded params are?

MikeCollingwood: Good point, yes! I had just manually updated them for the perf testing purpose but will raise another PR in core-common-infra

PR #1464, opened 2025-05-28
BAU: update spotless version

thebauSoftwire:
### What changed
Updated spotless to latest version

### Why did it change
The latest version hopefully resolves a [security alert that dependabot](https://github.com/govuk-one-login/ipv-stubs/security/dependabot/76) has raised as our current version of spotless doesn't support the patch for this alert.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,8 @@ repos:
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.5
     hooks:
+      - id: remove-tabs
+        exclude: ^([\w\/-]+)?build\.gradle$
       - id: remove-crlf
 
   - repo: https://github.com/mattlqx/pre-commit-search-and-replace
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -26,7 +26,6 @@ repos:
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.5
     hooks:
-      - id: remove-tabs
```
thebauSoftwire, line 29: this was conflicting with the new spotless version which wants to use tabs for the gradle build file but this removes it

DanCorderIPV: That's weird, we already have tabs in the gradle files. Maybe the question is why has this started affecting the gradle files?
I suspect we want to keep disallowing tabs elsewhere

thebauSoftwire: I'm not sure tbh 🤔  I know pre-commit only applies to updated files and the last few times this file was updated was with dependabot. If a dev updated it, maybe they hadn't had pre-commit installed and so it didn't apply?
But I've updated it now so that instead of removing the rule, I've added a file exclusion so that it ignores `build.gradle` files :)

DanCorderIPV (approved):

PR #1534, opened 2025-06-23
PYIC-8427: add new evcs identity endpoint

thebauSoftwire:
### What changed
- added new EVCS POST `/identity` endpoint as part of phase 1
-  re-use existing logic as the endpoint will be more like the existing PUT `/vcs` endpoint in phase two

### Why did it change
Trust & Reuse team have re-estimated the size of their work to instead now implement a POST `/identity` endpoint. As part of phase 1, we only need to use this endpoint to store the SI record. In phase 2, the endpoint will be updated to store the VCs as well.

### Issue tracking

- [PYIC-8427](https://govukverify.atlassian.net/browse/PYIC-8427)

## Checklists

[PYIC-8427]: https://govukverify.atlassian.net/browse/PYIC-8427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
+++ b/di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
@@ -19,6 +19,7 @@ import { v4 as uuid } from "uuid";
 import {
   EvcsItemForUpdate,
   PatchRequest,
+  PostIdentityRequest,
   PostRequest,
   PutRequest,
 } from "../domain/requests";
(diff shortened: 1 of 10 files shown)


di-ipv-evcs-stub/openAPI/evcs-external.yaml
```diff
@@ -384,6 +424,17 @@ components:
...
+      description: The request body for the /identity POST endpoint to store a user's stored identity record
+      type: object
+      properties:
+        userId:
+          type: string
+        si:
+          $ref: "#/components/schemas/StoredIdentityDetails"
+      required:
+        - userId
+        - si
```
thebauSoftwire, line 437: marking `si` as required here as it makes more sense since storing the SI is the primary function of this new endpoint in phase 1. In phase 2, where we will start using this endpoint to store VCs, this can be marked as not required again

di-ipv-evcs-stub/lambdas/src/services/evcsService.ts
```diff
@@ -76,89 +77,92 @@ export async function processPostUserVCsRequest(
   return response;
 }
 
-export async function processPutUserVCsRequest(
-  putRequest: PutRequest,
+export async function processPostIdentityRequest(
+  request: PutRequest | PostIdentityRequest,
 ): Promise<ServiceResponse> {
```
thebauSoftwire, line 82: re-using this function instead of creating a new one as the logic for the POST `identity` endpoint will eventually match the existing PUT endpoint in phase 2 so no need to delete in another PR

thebauSoftwire:
regression tested 🍏

blakeyp (approved):

PR #1671, opened 2025-08-28
PYIC-8381: create new SIS stub

thebauSoftwire:
### What changed

- addition of new SIS stub

**Testing notes:**
| case | notes|
|-|-|
| 200 | Used an access token generated by the api tests (I let a successful journey complete so that the associated userId had an SI record created by core-back) |
| 200 - Malformed JSON | Created an SI record via AWS console with missing vot/storedIdentity for a given user. Used the api tests to create an access token with the same userId. |
| 400 | Created an access token via the method above but removed the userId, missing request body, missing body.vtr, missing body.govukSigninJourneyId |
| 401 | Made a request without an auth header and with an incorrect bearer type |
| 403 | Made a request with a dummy access token |
| 404 | Used an access token generated by the api tests but for a scenario which wouldn't create an SI record) |
| 500 | deployed an instance of the stub which always returned a 500 |

### Why did it change

As part of phase 2 for SIS, we need to verify the stored identity record with the user's current VC bundle. To do this, we need to pull the SI record using a SIS endpoint. In lower environments, we need to interact with a stub instead.

### Issue tracking

- [PYIC-8381](https://govukverify.atlassian.net/browse/PYIC-8381)

## Checklists

[PYIC-8381]: https://govukverify.atlassian.net/browse/PYIC-8381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/sis-stub.yml
```diff
@@ -0,0 +1,66 @@
...
+      contents: read
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: '0'
+
+      - name: Set up AWS creds
+        uses: aws-actions/configure-aws-credentials@v4.3.1
+        with:
+          role-to-assume: ${{ secrets.SIS_STUB_BUILD_GH_ROLE }}
```
thebauSoftwire, line 29: not sure where to find these but this is also a reminder to set these before merging

thebauSoftwire: GH secrets have been added for SIS_STUB_BUILD_GH_ROLE and SIS_STUB_BUILD_S3_BUCKET

di-ipv-sis-stub/lambdas/src/services/sisService.ts
```diff
@@ -0,0 +1,55 @@
...
+    const { storedIdentity, levelOfConfidence, isValid } = unmarshall(siItem);
+
+    return {
+      content: storedIdentity,
+      vot: levelOfConfidence,
+      isValid,
+      // defaulting to false as the ttl is set to the default
+      // retention of VCs which is 120 years
+      expired: false,
+    };
```
blakeyp, line 43: There are a couple of extra fields `kidValid` and `signatureValid` which we can just default to true (they're new additions so not covered in the ticket sorry)

see https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/12bac6383cfe76b74f738cd4e6998cc86a9439f8/src/specs/api.yaml#L88

di-ipv-sis-stub/openAPI/sis-external.yaml
```diff
@@ -0,0 +1,170 @@
...
+        type: "MOCK"
+        requestTemplates:
+          application/json: "{\"statusCode\":200}"
+        responses:
+          200:
+            statusCode: 200
+            responseTemplates:
+              application/json: "{\"healthcheck\": \"ok\"}"
+  /user-identity:
+    get:
```
blakeyp: oo also sorry it's `POST` now, since they'd like us to pass the journey id with the request - see https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/12bac6383cfe76b74f738cd4e6998cc86a9439f8/src/specs/api.yaml#L59 for the shape of the request body

blakeyp (approved):

PR #1750, opened 2025-10-13
PYIC-8720: update sis stub response.content

thebauSoftwire:
### What changed
- Update SIS stub response to return the contents of the SI JWT and not the encoded JWT string
- Remove jwt signing methods that are no longer used

### Why did it change

According to the [SIS API schema](https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/main/openAPI/api.yaml#L117-L184), the response.content should be the contents of the SIS JWT and not the encoded JWT string itself.

### Issue tracking

- [PYIC-8720](https://govukverify.atlassian.net/browse/PYIC-8720)

## Checklists

[PYIC-8720]: https://govukverify.atlassian.net/browse/PYIC-8720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


di-ipv-sis-stub/lambdas/src/domain/userIdentity.ts
```diff
@@ -1,7 +1,7 @@
 import { JWTPayload } from "jose";
 
 export interface UserIdentity {
-  content: string;
+  content?: StoredIdentityContents;
```
thebauSoftwire, line 4: marking this as non-required as the ACs on the original [PYIC-8381](https://govukverify.atlassian.net/browse/PYIC-8381?search_id=77a87431-4333-4cd0-9936-1b06c8ca4a8c) says that it's possible to get a response which doesn't return a content field (ie when the stored identity record is missing required properties)

DanCorderIPV (commented):
All looks sensible to me, just waiting on confirmation that this is the right response shape

DanCorderIPV (approved):

PR #1809, opened 2025-11-05
SPT-1699: EVCS Added unsignedVot to GET /identity

chris-dco:
### What changed

Updated the EVCS `GET /identity` endpoint to return the additional field `unsignedVot` which will contain the value of the `vot` field within the DynamoDb record.

### Why did it change

There is been some confusion between the Vector of Trust (VOT) field within a signed Stored Identity JWT and the VOT field which is passed into EVCS when a Stored Identity is persisted. 

Currently EVCS and SIS are designed to return the VOT from within the signed Stored Identity JWT, however IPV Core were expecting the VOT which is stored alongside the signed JWT in a separate field in DynamoDb. This has lead to confusion as the VOT metadata may be `P3` but the Stored Identity JWT may be `P2`. So currently the Stored Identity Service are returning `P2` when IPV Core are expecting `P3`.

IPV Core are going to start including the `max_vot` field within the signed Stored Identity JWT which will be used by the Stored Identity Service moving forwards. However, there are a number of existing records where this field is not present. As a work around for this phase of work, SIS will use the `unsignedVot` when the `max_vot` is not present in the JWT.

### Issue tracking

- [SPT-1699](https://govukverify.atlassian.net/browse/SPT-1699)

## Checklists

[SPT-1699]: https://govukverify.atlassian.net/browse/SPT-1699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-evcs-stub/lambdas/src/domain/sharedTypes.ts
+++ b/di-ipv-evcs-stub/lambdas/src/domain/sharedTypes.ts
@@ -9,7 +9,7 @@ export interface VcDetails {
 
 export type StoredIdentityResponse = {
   si: Omit<VCIncludingStateAndMetadata, "state">;
-  vcs: VCIncludingStateAndMetadata[];
+  vcs: Omit<VCIncludingStateAndMetadata, "unsignedVot">[];
 };
 
 export interface VCIncludingStateAndMetadata {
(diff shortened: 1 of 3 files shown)


michaelabenyohai (approved):
As with my comments on the related [PR](https://github.com/govuk-one-login/ipv-identity-reuse-storage/pull/1653) I do think it would be so helpful to split the`VCIncludingStateAndMetadata` type into separate types for a VC and a stored identity. But approving as perhaps this is a refactor for a separate story.

PR #1836, opened 2025-11-21
PYIC-8804: resolve js-yaml vulnerability

thebauSoftwire:
### What changed
Manually update js-yaml to 4.1.1 in package-lock.json for cimit-stub-ts 

### Why did it change
Dependabot couldn't automatically create a PR as it was a transitive dependency

### Issue tracking

- [PYIC-8804](https://govukverify.atlassian.net/browse/PYIC-8804)

## Checklists

[PYIC-8804]: https://govukverify.atlassian.net/browse/PYIC-8804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-cimit-stub-ts/lambdas/package-lock.json
+++ b/di-ipv-cimit-stub-ts/lambdas/package-lock.json
@@ -2999,9 +2999,9 @@
       }
     },
     "node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-      "version": "3.14.1",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.14.1.tgz",
-      "integrity": "sha512-okMH7OXXJ7YrN9Ok3/SXrnu4iX9yOk+25nqX4imS2npuvTYDmo/QEZoqwZkYaIDk3jVvBOTOIEgEhaLOynBS9g==",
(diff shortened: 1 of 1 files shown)


undermad (commented):
Hi Thea, Is updating package-lock.json manually a good idea? As far as I know it may cause some issues and usually is updated by triggered by version change in package.json. I may be wrong ofc, but worth to double check that.

thebauSoftwire:
@undermad yeah it's not the most robust solution but this is a result of manually running `npm audit fix` which upgrades any vulnerable dependencies so that it's still compatible across the different dependencies. Hopefully will work 🤞 but will revert if not

undermad (approved):
Hmm, I think it can be done by running "npm ci" to check integrity between "package.json" and "package-lock.json". Also worth to double check if in deployment process we use "npm install" which may override "package-lock.json" file. 
Definitely worth to keep eye on it.

Edit: Sorry, It looks like "npm audit fix" does the job well actually. Just jumped into docs and it saying it try to resolve only compatible versions.
https://docs.npmjs.com/cli/v9/commands/npm-audit?utm_source=chatgpt.com

PR #1894, opened 2026-01-07
PYIC-8831: Update DCMAW Async Stub to simulate the use of an expired Drivers License

AmritSidhu:
### What changed

- Update DCMAW Async Stub to simulate the use of an expired Drivers License
- Add DL as a support document type and allow driving licence expiryDate to be overridden, this will allow us to dynamic test the expiryDate - expired, same day, future
- Instead of using a hard-coded expiry date (which means at some point the DL will be counted as expired), we can instead default to a DL VC which expires in future so we don't have to keep sending the request with a custom expiry date
- Extend the management enqueue VC schema to support driving permits and document the optional driving_permit_expiry_date field

### Why did it change

This allows teams to exercise and verify the new expired-licence grace-period logic and the midnight-pinned expiry rules in local and test environments.

### Issue tracking

- [PYIC-8831](https://govukverify.atlassian.net/browse/PYIC-8831)

## Checklists

[PYIC-8831]: https://govukverify.atlassian.net/browse/PYIC-8831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-dcmaw-async-stub/lambdas/src/data/vcEvidenceClaims.ts
+++ b/di-ipv-dcmaw-async-stub/lambdas/src/data/vcEvidenceClaims.ts
@@ -1,0 +1,73 @@
+import { DocumentType, EvidenceType } from "../domain/managementEnqueueRequest";
+
+export const EVIDENCE_CLAIMS = {
+  [DocumentType.ukChippedPassport]: {
+    [EvidenceType.success]: {
+      type: "IdentityCheck",
+      strengthScore: 4,
+      validityScore: 3,
+      checkDetails: [
+        {
+          checkMethod: "vcrypt",
+          identityCheckPolicy: "published",
+          activityFrom: null,
+        },
+        {
+          checkMethod: "bvr",
+          biometricVerificationProcessLevel: 3,
+        },
+      ],
+    },
+    [EvidenceType.fail]: {
+      type: "IdentityCheck",
+      strengthScore: 4,
+      validityScore: 0,
+      failedCheckDetails: [
+        {
+          checkMethod: "vcrypt",
+          identityCheckPolicy: "published",
+          activityFrom: null,
+        },
+        {
+          checkMethod: "bvr",
+          biometricVerificationProcessLevel: 3,
+        },
+      ],
+    },
+  },
+  [DocumentType.drivingPermit]: {
+    [EvidenceType.success]: {
+      type: "IdentityCheck",
+      strengthScore: 3,
+      validityScore: 2,
+      activityHistoryScore: 1,
+      checkDetails: [
+        {
+          checkMethod: "vcrypt",
+          identityCheckPolicy: "published",
+          activityFrom: null,
+        },
+        {
+          checkMethod: "bvr",
+          biometricVerificationProcessLevel: 3,
+        },
+      ],
+    },
+    [EvidenceType.fail]: {
+      type: "IdentityCheck",
+      strengthScore: 4,
+      validityScore: 0,
+      failedCheckDetails: [
+        {
+          checkMethod: "vcrypt",
+          identityCheckPolicy: "published",
+          activityFrom: null,
+        },
+        {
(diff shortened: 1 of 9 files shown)


di-ipv-dcmaw-async-stub/lambdas/src/data/vcEvidenceClaims.ts
```diff
@@ -0,0 +1,73 @@
...
+        },
+        {
+          checkMethod: "bvr",
+          biometricVerificationProcessLevel: 3,
+        },
+      ],
+    },
+    [EvidenceType.fail]: {
+      type: "IdentityCheck",
+      strengthScore: 4,
```
DanCorderIPV: This should be a 3?

di-ipv-dcmaw-async-stub/lambdas/src/domain/mockVc.ts
```diff
@@ -3,18 +3,41 @@ import {
...
+  const credentialSubject = {
+    ...USER_CLAIMS[testUser],
+    ...(isDrivingPermitCredentialSubject(documentDetails)
+      ? {
+          drivingPermit: [
+            {
+              ...documentDetails.drivingPermit[0],
+              expiryDate:
+                drivingPermitExpiryDate ??
+                getFutureExpiryDateStringFromIssuedAt(timestamp),
```
DanCorderIPV: `getFutureExpiryDateStringFromIssuedAt` seems a little misleading - there's no indication here that `timestamp` is the same as the `issuedAt` value

di-ipv-dcmaw-async-stub/lambdas/test/domain/mockVc.test.ts
```diff
@@ -0,0 +1,139 @@
+import { buildMockVc } from "../../src/domain/mockVc";
```
DanCorderIPV, line 1: Thanks for adding these

DanCorderIPV (changes requested):

di-ipv-dcmaw-async-stub/lambdas/src/domain/mockVc.ts
AmritSidhu: timestamp is already the single source of truth in this builder and is assigned to both iat and nbf in the returned VC.

The requirement for this change is that the default expiry is derived from the VC’s iat; using the same timestamp value ensures expiry stays anchored to the IssuedAt value we put in the VC.

Renaming the helper to something like ...FromTimestamp would actually be less accurate, because the timestamp here isn’t just any timestamp.

Could just assign timestamp to a variable called iat, not sure if thats much better.

@thebauSoftwire any ideas?

DanCorderIPV: I'm missing something then, where is it assigned to iat and nbf?

thebauSoftwire: hmm yeah I agree @DanCorderIPV that timestamp is quite poorly named and at the creation of the credentialSubject variable on like 25 doesn't have enough context around it to know that it will also be used for issued at. Though, I also agree with @AmritSidhu  that the function name conveys the intent. I think potentially moving it to within the `return {...}` block where you can see what the timestamp is used for (ie iat and nbf) more closely to when we pass it into `getFutureExpiryDateStringFromIssuedAt` would help with the giving context as to why we pass in the value to that function

DanCorderIPV: Oh, it's used in the final `return` - it was cut off by github, that makes much more sense!

DanCorderIPV (dismissed):

DanCorderIPV (approved):

PR #1925, opened 2026-01-16
PYIC-8879: update AIS stub to take custom state parameter

thebauSoftwire:
### What changed

- updated AIS management stub endpoint to take custom `state` parameter to override default provided by the `intervention`

### Why did it change

- There real AIS doesn't update the intervention type when the user performs a remedial action but instead updates the state block. Currently, in our tests, we prime the stub to return AIS_NO_INTERVENTION which is inaccurate. To make testing more accurate we need to be able to prime the AIS stub to return a custom state block.

### Issue tracking

- [PYIC-8879](https://govukverify.atlassian.net/browse/PYIC-8879)

## Checklists

[PYIC-8879]: https://govukverify.atlassian.net/browse/PYIC-8879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-ais-stub/README.md
+++ b/di-ipv-ais-stub/README.md
@@ -12,13 +12,20 @@ This endpoint takes a request body which references the intervention description
 {
   "statusCode": 200,
   "intervention": "AIS_NO_INTERVENTION",
-  "responseDelay": 0
+  "responseDelay": 0,
+  "state": {
+    "blocked": "false",
+    "suspended": "false",
+    "resetPassword": "false", // pragma: allowlist secret
+    "reproveIdentity": "false"
+  }
 }
 ```
 
 - `statusCode` set on the response.
 - `intervention` mapped to a response body in `/lambdas/src/data`.
 - `responseDelay` delays the response when called.
+- `state` can be optionally provided to override the default state block associated with an `intervention` in `/lambdas/src/data`
 
 ## AIS Endpoint
 
@@ -31,7 +38,7 @@ It returns a response with code and body primed by the management endpoint after
 In order to test it works, you can run the following on the deployed environment. It shows the call to the management endpoint to stub the account interventions, followed by the call to get account interventions. In this example, we are using the `dev-mikec` environment:
 
 ```bash
-curl \                                                                                                       
+curl \
    -d '{ "intervention": "AIS_NO_INTERVENTION" }' \
    -H "Content-Type: application/json" \
    -X POST https://ais-dev-mikec.02.core.dev.stubs.account.gov.uk/management/user/urn:uuid:42810001-5786-4a2d-a8ee-6d54cc386881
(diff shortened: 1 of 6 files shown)


di-ipv-ais-stub/README.md
```diff
@@ -12,13 +12,20 @@ This endpoint takes a request body which references the intervention description
 {
   "statusCode": 200,
   "intervention": "AIS_NO_INTERVENTION",
-  "responseDelay": 0
+  "responseDelay": 0,
+  "state": {
+    "blocked": "false",
+    "suspended": "false",
+    "resetPassword": "false", // pragma: allowlist secret
```
DanCorderIPV, line 19: Does `//` work in `.md` files?

di-ipv-ais-stub/lambdas/src/utils/dataLayer.ts
```diff
@@ -9,15 +9,18 @@ export default async function persistFutureAisResponse(
...
   console.info("Build response.");
   const response: Response = {
     userId: userId,
     statusCode: statusCode || 200,
     ttl: await getTtl(),
     responseDelay: responseDelay || 0,
-    responseBody: cases[intervention],
+    responseBody: {
+      ...cases[intervention],
+      ...(state ? {state} : {})
```
DanCorderIPV, line 22: If the use doesn't supply a state do we need to provide a more realistic default?

DanCorderIPV (changes requested):

di-ipv-ais-stub/lambdas/src/utils/dataLayer.ts
thebauSoftwire: by default, the state is provided by [cases[intervention]](https://github.com/govuk-one-login/ipv-stubs/blob/26ebefd27d7749281c082174ccf7cdff7a024452/di-ipv-ais-stub/lambdas/src/cases.ts#L2-L20) and the intervention is required unless providing a custom credentialSubject/evidence

di-ipv-ais-stub/README.md
thebauSoftwire: since it's in a code block and it rendered fine in intellij I assumed it was okay. I've removed the json specifier just in case it doesn't like that but without it it should render fine (it looks fine [here](https://github.com/govuk-one-login/ipv-stubs/tree/main/di-ipv-sis-stub))

DanCorderIPV: Ah, I thought that the pragma was needed to stop secret scanning complaining about this file

di-ipv-ais-stub/lambdas/src/utils/dataLayer.ts
DanCorderIPV: Ah, I misread that code as inserting empty state by default.

DanCorderIPV (approved):

di-ipv-ais-stub/README.md
thebauSoftwire: ah yeah it is, detect-secrets is able to pick up the pragma: allowlist secret. I thought the main concern was whether the README file would be rendered okay

PR #1937, opened 2026-01-21
PYIC-8830: update driving permit test data so that they always expire in the future

thebauSoftwire:
### What changed

- update driving permit test data to always have an expiry date in the future

Note: I've only updated the driving permits that are NOT used for staging as these require specific values that need to be valid for the DL CRI

### Why did it change

We're adding DL expiry checks in core-back which means that any tests which use driving permit details with hard-coded expiry dates will fail in the future. This commit should future-proof the tests so they won't need to be updated once the hard-coded dates pass.

### Issue tracking

- [PYIC-8830](https://govukverify.atlassian.net/browse/PYIC-8830)

## Checklists

[PYIC-8830]: https://govukverify.atlassian.net/browse/PYIC-8830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/data/CriStubDataDcmaw.java
+++ b/di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/data/CriStubDataDcmaw.java
@@ -16,135 +16,124 @@ public class CriStubDataDcmaw {
         // Replace default public constructor
     }
 
-    public static final List<IdentityCheckSubjectCriStubData> Data =
-            List.of(
-                    createData(
-                            "Alice Doe (Valid) NFC Passport",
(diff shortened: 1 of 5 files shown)


di-ipv-credential-issuer-stub/src/main/java/uk/gov/di/ipv/stub/cred/data/CriStubDataDcmaw.java
thebauSoftwire, line 1: looks like a big change but the main change is updating `Data` from a static variable to a static method `getData` so calling the method each time refreshes the expiry date on the driving permits (and calling the static methods wherever we use a driving permit)

AmritSidhu (approved):

PR #1939, opened 2026-01-22
PYIC-8758: Exclude generated dev-deploy templates from Checkov pre-commit scans

AmritSidhu:
### What changed

Configure the Checkov pre-commit hook to scan only deploy/**/template.yaml and explicitly skip dev-deploy and core-dev-deploy generated CloudFormation templates

### Why did it change

dev-deploy and core-dev-deploy generated CloudFormation templates that break linting due to stripped skip comments.

### Issue tracking

- [PYIC-8758](https://govukverify.atlassian.net/browse/PYIC-8758)

## Checklists

[PYIC-8758]: https://govukverify.atlassian.net/browse/PYIC-8758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,9 +54,11 @@ repos:
   - repo: https://github.com/bridgecrewio/checkov.git
     rev: '3.2.452'
     hooks:
-      - id: checkov
-        args: ['--framework', 'cloudformation', '--quiet']
-        files: ^(.*/)?template\.(yml|yaml)$
+      - id: checkov_diff
+        args:
+          - '--framework'
+          - 'cloudformation'
+          - '-f'
 
   - repo: https://github.com/lovesegfault/beautysh
     rev: 'v6.2.1'


.pre-commit-config.yaml
```diff
@@ -55,8 +55,16 @@ repos:
...
+        args:
+          - --framework
+          - cloudformation
+          - --quiet
+          - --skip-path
+          - core-dev-deploy
+          - --skip-path
+          - dev-deploy
+        files: ^(?!.*(?:^|/)(?:core-dev-deploy|dev-deploy)(?:/|$)).*/deploy/.*/template\.(yml|yaml)$
+
```
blakeyp: I think we only want to skip the dev-deploy generated files e.g. `orch-stub-patrickb.yaml` which dev-deploy copies from the templates (rather than the original templates themselves) ?

Also I've only seen this issue impacting the checkov check but the ticket mentions `cfn-python-lint` (the check above) so do we need to apply the same there too? cc @thebauSoftwire

thebauSoftwire: ah I think I actually did mean `checkov`, I wrote that the cause was dev-deploy stripping the `checkov:skip{RULE}` comments from the generated files. This was a mistake on my part, sorry! I'll update the ticket.

I think we would want the `cfn-python-lint` as it runs cloudformation validation checks?

blakeyp: I still think this will exclude templates we want to be scanned e.g. https://github.com/govuk-one-login/ipv-stubs/blob/main/di-ipv-orchestrator-stub/core-dev-deploy/template.yaml
We only want the dev-deploy generated files e.g. `core-dev-deploy/orch-stub-patrickb.yaml` to be excluded?

blakeyp (approved):
Nice 👍

PR #1942, opened 2026-01-28
Pyic 8757 stop deployment on test failure

DanCorderIPV:
### What changed

Run tests after merging to main but before deploying the changes

### Why did it change

In the case that incompatible PRs are both merged the problem will be caught before the changes are deployed anywhere.

### Issue tracking

- [PYIC-8757](https://govukverify.atlassian.net/browse/PYIC-8757)

[PYIC-8757]: https://govukverify.atlassian.net/browse/PYIC-8757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/cri-stub-test-build-and-push.yml
+++ b/.github/workflows/cri-stub-test-build-and-push.yml
@@ -18,7 +18,12 @@ on:
         required: true
 
 jobs:
+  test:
+    uses: ./.github/workflows/test-and-build.yml
+    permissions:
+      packages: read
   dockerBuildAndPush:
+    needs: test
     name: Docker build and push
     runs-on: ubuntu-22.04-arm
     timeout-minutes: 15
--- a/.github/workflows/test-and-build.yml
+++ b/.github/workflows/test-and-build.yml
@@ -102,3 +102,24 @@ jobs:
       - name: Test
         working-directory: ./di-ipv-ais-stub/lambdas
         run: npm run test
+  build-cimit-stub:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Setup .npmrc
+        working-directory: ./di-ipv-cimit-stub-ts/lambdas
+        run: |
+          cp .npmrc.template .npmrc && \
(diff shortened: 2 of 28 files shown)


.github/workflows/test-and-build.yml
```diff
@@ -0,0 +1,125 @@
+name: test and build
+
+on: workflow_call
```
undermad, line 3: This is nice reusable solution but the problem I see with it is that even if we make changes only to one stub (lets say ticf-stub or evcs-stub), deployment will depend on successful tests and build of all other stubs. This add lots of unnecessary computation.

You could add a inputs flags to this workflow and add if statement on the jobs - sudo code below:
```
workflow_call:
    inputs:
      ticf:
        type: boolean
        default: false
      evcs:
        type: boolean
        default: false
```

And a job:
```
build-ticf-stub:
  if: inputs.ticf
```

And then call it in each stub workflow with appropriate test flags:
```
test:
    uses: ./.github/workflows/test-and-build.yml
    with:
      ticf: true
```

undermad (commented):
Looking great but it could be optimised, assuming each stub works independently from the another.
Imagine if we make changes to each of the stub in one PR, then it will trigger 
((tests and build for all stubs) * (number of amended stubs)).

.github/workflows/cri-stub-test-build-and-push.yml
```diff
@@ -0,0 +1,121 @@
...
+      DYNATRACE_PAAS_TOKEN:
+        required: true
+      ECR_REPOSITORY:
+        required: true
+      GH_ACTIONS_ROLE_ARN:
+        required: true
+
+jobs:
+  test:
+    uses: ./.github/workflows/test-and-build.yml
```
undermad, line 22: I don't think this is required here as it will test and build other irrelevant stubs?

.github/workflows/test-and-build.yml
DanCorderIPV: Good point, I didn't think this through when adding the TS stubs. I'll re-arrange this

.github/workflows/cri-stub-test-build-and-push.yml
DanCorderIPV: It shouldn't test the TypeScript stubs so I'll change that, but it does need to be here as that's the point of this ticket. We want to not deploy if the tests fail, so this step has to be in here so the steps below can use `needs: test`

DanCorderIPV:
This PR has been superceded

PR #2030, opened 2026-03-16
PYIC-8974: migrate CIMIT stub to Vitest/ESM

AmritSidhu:
### What changed

- migrate EVCS stub tests to Vitest
- migrate EVCS stub to native ESM
- remove unnecessary jest dependencies and config (including babel) upgrade eslint to >v9
- upgrade jose to > v6
- migrated your test suite to Vitest by replacing the strict, "official" AWS types with minimal local interfaces and using an unknown type assertion bridge to satisfy both the TypeScript compiler and ESLint rules without using any.

### Why did it change

To ensure continued security support for the jose library which we use in our stubs, we need to upgrade to jose v6 (ESM). This requires a full migration of each stub to a native ESM environment and a switch from Jest to Vitest for testing compatibility.

### Issue tracking

- [PYIC-8974](https://govukverify.atlassian.net/browse/PYIC-8974)

## Checklists

[PYIC-8974]: https://govukverify.atlassian.net/browse/PYIC-8974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanCorderIPV:
Why don't we want to use the official types? This way our tests will continue to pass even if the real types change in incompatible ways.

AmritSidhu:
> Why don't we want to use the official types? This way our tests will continue to pass even if the real types change in incompatible ways.

I think its the opposite.

In these tests, the AWS types require a object with loads of properties just to pass. By using local interfaces and casting, we only define the body and pathParameters the code actually reads. This keeps the tests focused on the logic and stops them from failing if AWS updates a property we don't even use.

thebauSoftwire:
I'm leaning towards using the real types because we'd lose the type strictness of our tests which means if the actual types change (e.g. update in field name, something becomes required instead of optional etc), since we're casting to our own type, the tests might pass and TS won't throw but at runtime, the code would fail. We'd need to keep our `MockQueryResponse` up to date ourselves to prevent this which we could easily forget.

Instead of mocking it out manually with vitest, I think the better approach would be to align how we mock dynamodb in our other stubs e.g. in the EVCS stub using [aws-sdk-client-mock](https://www.npmjs.com/package/aws-sdk-client-mock#install) and [aws-sdk-client-mock-vitest](https://www.npmjs.com/package/aws-sdk-client-mock-vitest) which handles a lot of the knotty type stuff for us. [Here's an example](https://github.com/govuk-one-login/ipv-stubs/blob/a041f97a606cd23f13cd76464105e3b5eaa63f1b/di-ipv-evcs-stub/lambdas/test/services/evcsService.test.ts#L108).

It might involve quite a bit of a refactor for the tests though so we could do it as a separate tech debt ticket if you feel it's too big for this work since this is blocking something else.

di-ipv-cimit-stub/lambdas/test/common/cimitStubItemService.test.ts
```diff
@@ -1,152 +1,169 @@
...
+      expect(dbMock).toHaveReceivedCommand(PutItemCommand);
+
+      const callArgs = dbMock.commandCalls(PutItemCommand)[0].args[0].input;
+      expect(callArgs.TableName).toBe("mock-table");
+      expect(callArgs.Item).toBeDefined();
+
+      // @ts-expect-error - Accessing marshalled DynamoDB attribute structure
+      expect(callArgs.Item.ttl.N).not.toBe("0");
+      // @ts-expect-error - Accessing marshalled DynamoDB attribute structure
+      expect(callArgs.Item.sortKey.S).toBe(ciCode + "#" + issuanceDate);
```
thebauSoftwire: you can use `unmarshall(callArgs.Item)` to turn the DynamoDB object into a JS object to avoid these error suppressions. Something like this could work:
```
expect(callArgs.Item).toBeDefined();
  const unmarshalledItem = unmarshall(
    callArgs.Item as Record<string, AttributeValue>,
  );
  expect(unmarshalledItem.ttl).not.toBe("0");
  expect(unmarshalledItem.sortKey).toBe("D01#2023-08-17T10:20:53.000Z");
```

di-ipv-cimit-stub/lambdas/test/common/cimitStubItemService.test.ts
```diff
@@ -1,152 +1,169 @@
...
-    expect(callArgs.TableName).toBe("mock-table");
-    expect(callArgs.Item).toBeDefined();
-    expect(callArgs.Item.ttl.N).not.toBe("0");
-    expect(callArgs.Item.sortKey.S).toBe("D01#2023-08-17T10:20:53.000Z");
+      const callArgs = dbMock.commandCalls(PutItemCommand)[0].args[0].input;
+
+      // @ts-expect-error - Accessing marshalled DynamoDB attribute structure
+      expect(callArgs.Item.ttl.N).not.toBe("0");
+      // @ts-expect-error - Accessing marshalled DynamoDB attribute structure
+      expect(callArgs.Item.sortKey.S).toBe("D01#2023-08-17T10:20:53.000Z");
```
thebauSoftwire: same comment as above

di-ipv-cimit-stub/lambdas/test/common/preMitigationService.test.ts
```diff
@@ -42,22 +44,25 @@ describe("preMitigationService", () => {
   });
 
   describe("applyPreMitigationsToItems", () => {
+    const createCimitItem = (
```
thebauSoftwire, line 47: nice!

di-ipv-cimit-stub/lambdas/test/external-api/stub-management/stubManagementHandler.test.ts
```diff
@@ -248,13 +264,11 @@ describe("stubManagementHandler", () => {
...
       });
-    });
 
-    ["POST", "PUT"].forEach((method) => {
       it(`should return 404 if CI item not found for mitigation request - ${method}`, async () => {
-        (
-          cimitStubItemService.getCiForUserId as jest.Mock
-        ).mockResolvedValueOnce(null);
+        vi.mocked(cimitStubItemService.getCiForUserId).mockResolvedValueOnce(
+          [],
```
thebauSoftwire, line 270: good spot!

di-ipv-cimit-stub/lambdas/test/external-api/stub-management/stubManagementHandler.test.ts
```diff
@@ -267,18 +281,12 @@ describe("stubManagementHandler", () => {
 
         expect(response.statusCode).toBe(404);
         expect(response.body).toContain("not found");
-        expect(cimitStubItemService.getCiForUserId).toHaveBeenCalledWith(
-          "123",
-          "456",
-        );
```
thebauSoftwire, line 273: do we not need this assertion anymore?

thebauSoftwire (commented):
nice! 🎉 Just a couple of small comments and a question

thebauSoftwire (approved):

PR #2064, opened 2026-04-01
PYIC-8458: Update /identity POST endpoint to accept vcs.

undermad:
### What changed

- POST /identity endpoint now accepts vcs array and govuk_signin_journey_id
- vc can be updated according to the allowed state transitions, otherwise will be rejected with 409
- vc can be added with any state
- vcs are validated
- si is validated

### Why did it change

In phase 1, a POST /identity endpoint was added to the EVCS stub which accepts a stored identity object to be stored in EVCS. 

To align with the EVCS API updates, we need to update the POST /identity endpoint to accept a VCs list so they can be saved to EVCS. This will be used to replace the existing POST /vcs/<userId> endpoint where we need to create an SI record and create/update a user’s VCs.

The api spec can be found [here](https://github.com/govuk-one-login/ipv-identity-reuse-storage/blob/7b551a6d0a0c3735988af5cc63045aa459324ca5/src/specs/api.yaml#L512-L565).

### Issue tracking

- [PYIC-8458](https://govukverify.atlassian.net/browse/PYIC-8458)

## Checklists

[PYIC-8458]: https://govukverify.atlassian.net/browse/PYIC-8458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/di-ipv-evcs-stub/lambdas/src/common/utils.ts
+++ b/di-ipv-evcs-stub/lambdas/src/common/utils.ts
@@ -1,4 +1,39 @@
+import { decodeJwt } from "jose";
+
 export function getErrorMessage(error: unknown) {
   if (error instanceof Error) return error.message;
   return String(error);
+}
+
+export function isValidJWT(token: string): boolean {
+  if (!token) {
+    return false;
+  }
+  const parts = token.split(".");
+  if (parts.length !== 3) {
+    return false;
+  }
+  const [header, payload, signature] = parts;
+  if (!header || !payload || !signature) {
+    return false;
+  }
+  const isBase64Url = (str: string) => /^[A-Za-z0-9_-]+$/.test(str);
+  if (
+    !isBase64Url(header) ||
+    !isBase64Url(payload) ||
+    !isBase64Url(signature)
+  ) {
+    return false;
+  }
+
+  try {
+    decodeJwt(token);
+  } catch {
+    return false;
+  }
+  return true;
+}
+
+export function getSignatureFromJwt(jwt: string): string {
+  return jwt.split(".")[2];
 }
--- a/di-ipv-evcs-stub/lambdas/src/common/utils.ts
+++ b/di-ipv-evcs-stub/lambdas/src/common/utils.ts
@@ -1,4 +1,5 @@
 import { decodeJwt } from "jose";
+import { decode } from "jose/base64url";
 
 export function getErrorMessage(error: unknown) {
   if (error instanceof Error) return error.message;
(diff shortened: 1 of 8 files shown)


undermad:
Note that 409 response is not configured in original EVCS service for this endpoint but it is in the ticket specification.

di-ipv-evcs-stub/lambdas/src/common/utils.ts
```diff
@@ -1,4 +1,39 @@
...
+  }
+  const parts = token.split(".");
+  if (parts.length !== 3) {
+    return false;
+  }
+  const [header, payload, signature] = parts;
+  if (!header || !payload || !signature) {
+    return false;
+  }
+  const isBase64Url = (str: string) => /^[A-Za-z0-9_-]+$/.test(str);
```
blakeyp: Might as well declare outside the function?

di-ipv-evcs-stub/lambdas/src/common/utils.ts
```diff
@@ -1,4 +1,39 @@
...
+  if (
+    !isBase64Url(header) ||
+    !isBase64Url(payload) ||
+    !isBase64Url(signature)
+  ) {
+    return false;
+  }
+
+  try {
+    decodeJwt(token);
```
blakeyp: This probably covers a lot of the steps above, just wondering if they're really necessary?

di-ipv-evcs-stub/lambdas/src/domain/requests/identityEndpointRequests.ts
```diff
@@ -8,16 +8,13 @@ export interface StoredIdentityDetails {
   expired?: boolean; // For use through the management API
 }
 
-export type PostIdentityRequest = Omit<PutRequest, "vcs" | "si"> & {
+export type PostIdentityRequest = {
+  userId: string;
+  govuk_signin_journey_id?: string;
+  vcs?: VcDetails[];
```
blakeyp, line 14: When only the new endpoint is being used we'll want to make this required (and validate it) - maybe add a note to [PYIC-9011](https://govukverify.atlassian.net/browse/PYIC-9011) ?

di-ipv-evcs-stub/lambdas/src/common/utils.ts
undermad: I moved declaration to its own scope. Also, used `decode()` function from `jose/base64url` to validate the format. This should be more accurate and handle all edge cases.
You can view the implementation [here](https://github.com/panva/jose/blob/main/src/util/base64url.ts).

di-ipv-evcs-stub/lambdas/src/common/utils.ts
undermad: This function validates only shape of the payload part in JWT. It also check if the token has 3 parts after split using "."
I rearranged the order little bit to remove duplicated validations. 
You can view the implementation [here](https://github.com/panva/jose/blob/main/src/util/decode_jwt.ts).

di-ipv-evcs-stub/lambdas/src/domain/requests/identityEndpointRequests.ts
undermad: According to this [specification](https://github.com/govuk-one-login/ipv-identity-reuse-storage/blob/7b551a6d0a0c3735988af5cc63045aa459324ca5/src/specs/api.yaml#L822-L834), only si is mandatory to be present in the request body. 
Not sure if that will change in the future?

blakeyp: You're right it will still technically be optional so we can leave this

blakeyp (approved):

di-ipv-evcs-stub/lambdas/src/common/utils.ts
```diff
@@ -1,4 +1,36 @@
...
+  } catch {
+    return false;
+  }
+  const parts = token.split(".");
+  const header = parts[0];
+  const signature = parts[2];
+  return !(!isBase64Url(header) || !isBase64Url(signature));
+}
+
+export function isBase64Url(str: string): boolean {
```
blakeyp, line 25: Not a big deal but doesn't need to be exported if not used outside of this file?

undermad: Indeed not needed here, but in this case it lives in utils file. I assuming we want to export utility function to be usable outside even if not yet used. - Maybe this is incorrect but I though for a fraction of the second this may be useful. I wouldn't do similar with other files/functions.

